Issue #8, opened 2025-03-04
Create graph representation of tram rail network geometry


RCRalph:
### Description

In order to represent realistic geometry of tram rail network, we need to utilize OpenStreetMap data which contains very precise information about rail geometry and location all around the city. This dataset isn't perfect for our use-case, because the distances between nodes may vary greatly. Our job is to make it fit our use-case, by following a few rules:

1. The placement of tram stops and intersections cannot be changed, the coordinates of them must remain as they are.
2. Each path must start and end at an intersection or a tram stop, with an exception for rails which are not continuous (for example the side track on *Cichy Kącik* or unfinished rail near *Jarzębiny*).
3. Each path must have equally spaced nodes. The node count should be as small as possible. The max distance between nodes cannot be greater than a given parameter value in meters. In my mind I was thinking of 1, 2, 5, or 10 meter spacing, but please develop the code in a way which would allow any value (also float), so it's more general and easier to change later if needed.
4. If any newly created nodes appear during the processing of the data, they should lay on the path marked by original OSM data as we don't want to use any interpolation techniques just yet. For example:
   ![Image](https://github.com/user-attachments/assets/e79c67cf-a220-4141-81fd-131e5bfd1f26)
   black - original OSM data, red - newly created nodes

**Examples of correct transformation:**  
![Image](https://github.com/user-attachments/assets/7da21d22-e41e-4fbd-9aca-6b875a8648a5) 
![Image](https://github.com/user-attachments/assets/ca985fd8-08bf-4d40-9449-09fbe0d1bdc1)

During the process of transforming the OSM data, we would also like to compute the distance between each node and the azimuth of each segment (remember that we're working with rectangular coordinates on a sphere, so please use `geopy` or any other library like that for precise results). As for the IDs of newly created nodes, please give them a unique positive integer value which doesn't conflict with any tram stop or intersection. The transformed graph should be a directed graph, but should allow bi-directional travel where possible (take into account the `oneway` tag value for each OSM way).

### Definition of done
We have a script which generates a simplified graph of tram rail network.
The script should work for any supplied city configuration.
For ease of testing, the configuration should be supplied as an argument.

PR #17, opened 2025-03-27
Added tram railway network graph transformer

Redor144:
### Related issues
- #8 

### Short description
Added `NetworkGraph` class to build a directed tram network graph from OpenStreetMap data. Includes functionality for subdividing track geometries and visualizing the graph on a map.


RCRalph (changes requested):
First round of review, probably out of many :3

server/src/overpass_client/overpass_client.py
```diff
@@ -25,6 +25,36 @@ class OverpassClient:
...
+
+    _GET_TRAM_STOPs_AND_CROSSROADS_COORDINATE = """
+    [out:json];
+    area["name"="{area_name}"]->.search_area;
+    way["railway"="tram"](area.search_area)->.tram_ways;
+    node(w.tram_ways);
+    out geom;
+    """
+
+    _GET_TRACK_GEOMETRY = """
```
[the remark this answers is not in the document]
RCRalph: I guess it should stay here, but not for long :3

server/src/graph_maker/create_graph.py
```diff
@@ -0,0 +1,357 @@
...
+      consecutive nodes does not exceed the specified max_distance parameter.
+    """
+
+    def __init__(
+        self,
+        ways_data: list[Way],
+        stops_data: list[TramStop],
+        coords_data: list[NodeCoordinate],
+        geometry_data: TrackGeometry,
+        max_distance=25.0,
```
RCRalph, line 24: Let's maybe change the constructor parameters to include only:
- Query result `tram_stops_and_tracks` of type `overpy.Result`
- `max_distance_in_meters` of type `float` without a default value
 
Please add a check if the float parameter is positive, otherwise raise a `ValueError` with appropriate message. Please perform the necessary changes to other methods in this class.

server/src/graph_maker/create_graph.py
```diff
@@ -0,0 +1,452 @@
...
+            with open(stops_file, "r", encoding="utf-8") as f:
+                self.stops = json.load(f)
+        except (FileNotFoundError, json.JSONDecodeError) as e:
+            raise ValueError(f"Error loading files: {e}")
+        self.geometry_file = geometry_file
+        self.coords_file = coords_file
+        self.max_distance = max_distance
+        self.main_nodes = []
+
+    def find_main_nodes(self):
```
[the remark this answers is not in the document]
RCRalph: The comment explains it now, thanks for adding it

server/src/model/osm_data.py
```diff
@@ -0,0 +1,33 @@
...
+
+class TrackGeometryFeature(BaseModel):
+    type: str
+    properties: Dict
+    geometry: Dict
+
+
+class TrackGeometry(BaseModel):
+    type: str
+    features: List[TrackGeometryFeature]
```
RCRalph, line 33: This looks like you got highly inspired by the structure of geojson. I would say we don't need inspiration from this file format and we should be only working with data structures provided by `overpy` to simplify the implementation.

server/src/graph_maker/create_graph.py
```diff
@@ -0,0 +1,452 @@
...
+        self, ways_file, stops_file, coords_file, geometry_file, max_distance=25.0
+    ):
+        self.graph = nx.DiGraph()
+        try:
+            with open(ways_file, "r", encoding="utf-8") as f:
+                self.ways = json.load(f)
+            with open(stops_file, "r", encoding="utf-8") as f:
+                self.stops = json.load(f)
+        except (FileNotFoundError, json.JSONDecodeError) as e:
+            raise ValueError(f"Error loading files: {e}")
```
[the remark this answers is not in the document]
RCRalph: Constructor parameters have been changed.

server/src/graph_maker/create_graph.py
```diff
@@ -0,0 +1,452 @@
...
+        self.geometry_file = geometry_file
+        self.coords_file = coords_file
+        self.max_distance = max_distance
+        self.main_nodes = []
```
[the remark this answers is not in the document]
RCRalph: Added protected properties

server/src/graph_maker/create_graph.py
```diff
@@ -0,0 +1,452 @@
...
+
+g = NetworkGraph(
+    "tram_ways.json",
+    "tram_stops.json",
+    "tram_stops_crossroads_coords.json",
+    "track_geometry.json",
+    max_distance=25,
+)
+g.create_graph()
+g.visualize_graph(map_filename="graph_map_krk.html")
```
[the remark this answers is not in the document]
RCRalph: Removed the code

server/src/graph_maker/create_graph.py
```diff
@@ -0,0 +1,452 @@
...
+import math
+
+import folium
+import networkx as nx
+import pyproj
+from shapely.geometry import LineString, MultiLineString, Point
+from shapely.ops import linemerge, transform
+
+
+class NetworkGraph:
```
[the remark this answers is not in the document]
RCRalph: Changed the name

server/src/graph_maker/create_graph.py
```diff
@@ -0,0 +1,452 @@
...
+    *is_there_any_ways_to_merge method checks if there are any ways to merge
+    *merge_ways method merges ways that are close to each other:
+        -we are iterating through ways and checking if there are ways that are close to each other
+        -if there are we are merging them and updating nodes
+    *update_nodes_after_merge method updates nodes after merge
+    *subdivide_ways method divides ways into smaller parts:
+        -we are iterating through ways and checking if there are nodes in the way
+        -if there are we are dividing way into smaller parts
+    *visualize_graph method visualizes the graph on the map
+    """
```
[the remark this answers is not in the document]
RCRalph: Changed the docstring

server/src/graph_maker/create_graph.py
```diff
@@ -0,0 +1,452 @@
...
+                folium.CircleMarker(
+                    location=(lat, lon),
+                    radius=3,
+                    color="blue",
+                    fill=True,
+                    fill_opacity=0.6,
+                    popup=f"New node: {node}",
+                ).add_to(fmap)
+        fmap.save(map_filename)
+        print(f"Map saved as: {map_filename}")
```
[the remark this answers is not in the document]
RCRalph: Removed the method.

server/src/model/osm_data.py
```diff
@@ -0,0 +1,33 @@
...
+    id: int
+    lat: float
+    lon: float
+    name: Optional[str] = None
+
+
+class NodeCoordinate(BaseModel):
+    id: int
+    lat: float
+    lon: float
```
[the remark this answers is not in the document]
olobuszolo: You're right. These classes don't make much sense, so I've deleted the file. We'll work directly with overpy.Result instead.

server/src/graph_maker/create_graph.py
olobuszolo: I've updated the constructor according to your suggestion and added a check to ensure that `max_distance_in_meters` is positive. Also now `max_distance_in_meters` does not have default value.

server/src/model/osm_data.py
olobuszolo: You're right — these classes were inspired by geojson, but they added unnecessary complexity. They've now been deleted.

server/src/overpass_client/overpass_client.py
```diff
@@ -25,6 +25,36 @@ class OverpassClient:
...
+    area["name"="{area_name}"]->.search_area;
+    (
+    way["railway"="tram"](area.search_area);
+    relation["railway"="tram"](area.search_area);
+    );
+    out body;
+    >;
+    out skel qt;
+    """
+
```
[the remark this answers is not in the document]
olobuszolo: Good catch- your trick works perfectly and has been applied. Props to the secret source

server/src/overpass_client/overpass_client.py
```diff
@@ -36,3 +66,18 @@ def get_relations_and_stops(cls, area_name: str, custom_node_ids: list[int]):
...
+    @classmethod
+    def get_tram_stops_and_crossroads_coordinate(cls, area_name: str):
+        query = cls._GET_TRAM_STOPs_AND_RAILWAYS_COORDINATE.format(area_name=area_name)
+        return cls._OVERPASS.query(query)
+
+    @classmethod
+    def get_track_geometry(cls, area_name: str):
+        query = cls._GET_TRACK_GEOMETRY.format(area_name=area_name)
+        return cls._OVERPASS.query(query)
```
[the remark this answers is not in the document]
olobuszolo: You're right here as well — I've applied your suggestion.

server/src/graph_maker/city_data_downloader.py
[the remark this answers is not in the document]
olobuszolo: You're right — this class, in its current form, is unnecessary. I've removed it

server/src/graph_maker/create_graph.py
```diff
@@ -0,0 +1,375 @@
+import math
+
+import networkx as nx
+import overpy
+import pyproj
+from shapely.geometry import LineString, MultiLineString, Point
+from shapely.ops import linemerge, transform
+
+
+class TramRailwayGraphTransformer:
```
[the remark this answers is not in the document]
olobuszolo: This new name is better. I've changed the name of class and updated the name of the file.

server/src/graph_maker/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .tram_truck_graph_transformer import TramTrackGraphTransformer
```
RCRalph, line 1:
```suggestion
from .tram_track_graph_transformer import TramTrackGraphTransformer
```

server/src/graph_maker/create_graph.py
```diff
@@ -0,0 +1,375 @@
...
+    """
+    TramRailwayGraphTransformer generates a directed graph of the tram and railway network in two stages:
+    - First, it builds a skeleton graph using only main nodes (tram stops or tram crossroads).
+    - Then, it densifies the graph by adding intermediate nodes between main nodes so that the distance between
+      consecutive nodes does not exceed the specified max_distance parameter.
+    """
```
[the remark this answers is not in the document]
olobuszolo: Okay, so I’ve renamed main nodes to `permanent` nodes as suggested. I’ve also marked methods as protected. The graph creation is now split into two separate stages as you proposed: first, we build the skeleton (which is stored as a separate graph), and then we densify it.

server/src/graph_maker/__init__.py
olobuszolo: Sure, updated file name

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,366 @@
...
+
+    def densify_graph_by_max_distance(
+        self, max_distance_in_meters: float
+    ) -> nx.DiGraph:
+        if max_distance_in_meters <= 0:
+            raise ValueError("max_distance_in_meters must be a positive float value")
+
+        graph_copy = self._graph.copy()
+        self._subdivide_ways(graph_copy, max_distance=max_distance_in_meters)
+        return graph_copy
```
RCRalph, line 62: I don't know how about you, but personally I'm a fan of ordering the methods in a class or a file wherever possible in such a way, where no method of the class calls a method below itself (with the exception of `__init__` as it's the class constructor so we expect it to be at the top), for example:

```py
class A:
    def __init__(self):
        self.x()
        self.y()

    def z(self):
        pass

    def x(self):
        self.z()

    def y(self):
        self.z()
        self.x()
```

Following this principle, these two methods should be at the bottom of the class as they require an already constructed class instance, while the methods used in the constructor should be just below it. What do you think?

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,366 @@
...
+    def __init__(self, tram_stops_and_tracks: overpy.Result):
+        self._ways = tram_stops_and_tracks.ways
+        self._stops = [
+            node
+            for node in tram_stops_and_tracks.nodes
+            if node.tags.get("railway") == "tram_stop"
+        ]
+        self._coords_data = tram_stops_and_tracks.nodes
+        self._graph = nx.DiGraph()
+        self._pernament_nodes = []
```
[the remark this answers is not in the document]
olobuszolo: Fixed typo and changed `_permanent_nodes` to 'set'.

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,366 @@
...
+        return graph_copy
+
+    def _find_pernament_nodes(self):
+        """
+        Identifies main nodes in the _graph — nodes representing tram stops or railway crossings.
+        These nodes form the basis for the initial _graph skeleton.
+        They are essential for filtering out unnecessary nodes before the second stage of _graph creation,
+        which involves densifying the network between main nodes.
+        """
+        result = [
```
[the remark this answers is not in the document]
olobuszolo: Okay, that's great suggestion. Now method `_find_permanent_nodes` is calling methods: `_find_permanent_crossing` and `_find_permanent_tram_stops` and return union of them

server/src/graph_maker/tram_track_graph_transformer.py
olobuszolo: You're right. I also changed the order of `_update_nodes_after_merge` and `_merge_ways` and moved `_initialize_graph` to the end of the file but maybe this could be exception? It actually shows the subsequent stages of creating the graph. What do you think?

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,366 @@
...
+        self._pernament_nodes = []
+        self._ways_dict = {}
+
+        self._build_skeleton_graph()
+
+    def _build_skeleton_graph(self):
+        for way in self._ways:
+            node_ids = [node.id for node in way.nodes]
+            for i in range(len(node_ids) - 1):
+                self._graph.add_edge(node_ids[i], node_ids[i + 1])
```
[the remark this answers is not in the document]
olobuszolo: I extracted this into a separate method called `_build_initial_graph`. The other methods that further construct the graph are grouped within `_initialize_graph`.
Two-way tracks must be handled in a separate method after removing unnecessary nodes, because the graph is first built and then `_permanent_nodes` are identified.
Adding two-way edges before cleaning up unnecessary nodes breaks the logic for detecting crossings — a regular node that's not a real crossing but belongs to a two-way edge ends up with multiple incoming/outgoing edges. In this graph-building approach, that would incorrectly classify it as a crossing.

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,386 @@
...
+            stop.id
+            for stop in self._stops
+            if stop.id in self._graph.nodes
+            and (
+                self._graph.out_degree(stop.id) > 0
+                or self._graph.in_degree(stop.id) > 0
+            )
+        )
+
+        return result
```
RCRalph, line 61: You can use a set comprehension here. Also there's no need to assign it to variable `result` as it's returned directly in the next instruction.

```suggestion
        return {
            stop.id
            for stop in self._stops
            if stop.id in self._graph.nodes
            and (
                self._graph.out_degree(stop.id) > 0
                or self._graph.in_degree(stop.id) > 0
            )
        }
```

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,386 @@
...
+
+    def __init__(self, tram_stops_and_tracks: overpy.Result):
+        self._ways = tram_stops_and_tracks.ways
+        self._stops = [
+            node
+            for node in tram_stops_and_tracks.nodes
+            if node.tags.get("railway") == "tram_stop"
+        ]
+        self._coords_data = tram_stops_and_tracks.nodes
+        self._permanent_nodes = set()
```
RCRalph, line 29: Please add a type hint so it's obvious what this set actually contains

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,386 @@
...
+
+        return result
+
+    def _find_permanent_crossing(self):
+        result = set()
+        for node in self._graph.nodes:
+            if self._graph.in_degree(node) > 1 or self._graph.out_degree(node) > 1:
+                result.add(node)
+
+        return result
```
RCRalph, line 69: Also set comprehension here is a good idea
```suggestion
        return {
            node
            for node in self._graph.nodes
            if self._graph.in_degree(node) > 1 or self._graph.out_degree(node) > 1
        }
```

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,386 @@
...
+
+        for way in self._ways:
+            node_ids = [node.id for node in way.nodes]
+
+            for i in range(len(node_ids) - 1):
+                graph.add_edge(node_ids[i], node_ids[i + 1])
+
+        return graph
+
+    def _find_permanent_tram_stops(self):
```
RCRalph, line 50: I think a better name for this method would be `_get_tram_stop_node_ids_in_graph`, because at this level we don't particularly care that these nodes are permanent and it's also a good idea to tell the user that the result will contain IDs of nodes.

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,386 @@
...
+            if stop.id in self._graph.nodes
+            and (
+                self._graph.out_degree(stop.id) > 0
+                or self._graph.in_degree(stop.id) > 0
+            )
+        )
+
+        return result
+
+    def _find_permanent_crossing(self):
```
RCRalph, line 63: Similar to comment above, `_get_crossing_node_ids_in_graph` would be a better name, what do you think?

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,386 @@
...
+        result = set()
+        for node in self._graph.nodes:
+            if self._graph.in_degree(node) > 1 or self._graph.out_degree(node) > 1:
+                result.add(node)
+
+        return result
+
+    def _find_permanent_nodes(self):
+        """
+        Identifies main nodes in the _graph — nodes representing tram stops or railway crossings.
```
RCRalph, line 73: This sentence is unnecessary as now it's clear what nodes are permanent from the code itself.
Can you see that we just made this method much more readable than it was before the code review?

server/src/graph_maker/tram_track_graph_transformer.py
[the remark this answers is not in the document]
RCRalph: Great, now put this explanation as a comment in the code so it's clear for everyone. I suggest adding it to the docstring of `_build_initial_graph` as a mention of two-way tracks is already there.

server/src/graph_maker/tram_track_graph_transformer.py
RCRalph: I think it should be placed one method above as `_subdivide_ways` method is used only in `densify_graph_by_max_distance`. You can also change the order to how these methods are executed in the code, it's just a developer preference. In the example I've provided above it would look like this:
```py
class A:
    def __init__(self):
        self.x()
        self.y()
    
    def x(self):
        self.z()

    def z(self):
        pass

    def y(self):
        self.z()
        self.x()
```

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,386 @@
...
+        if max_distance_in_meters <= 0:
+            raise ValueError("max_distance_in_meters must be a positive float value")
+
+        graph_copy = self._graph.copy()
+        self._subdivide_ways(graph_copy, max_distance=max_distance_in_meters)
+        return graph_copy
+
+    @property
+    def permanent_node_ids(self):
+        return list(self._permanent_nodes)
```
RCRalph, line 386: I think this way is more explicit as to what is going on here, also there's no shame in returning a set instead of a list.
```suggestion
        return self._permanent_nodes.copy()
```

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,386 @@
...
+    TramTrackGraphTransformer converts OSM data into a directed NetworkX graph in two main stages:
+
+        1. Skeleton graph construction:
+        - Builds an initial graph (_graph) using only main nodes — tram stops and tram track intersections.
+
+        2. Graph densification:
+        - Using the `densify_graph_by_max_distance` method, the graph is densified by adding  intermediate nodes so that the distance between consecutive nodes does not exceed the specified threshold.
+    """
+
+    def __init__(self, tram_stops_and_tracks: overpy.Result):
```
RCRalph, line 21: We're missing the `_graph` class attribute in the constructor. As you probably know, it's generally a bad practice to introduce class attributes outside of the constructor. The name of this variable could also be changed to `_tram_track_graph` which is more precise, what do you think?

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,386 @@
...
+        self._stops = [
+            node
+            for node in tram_stops_and_tracks.nodes
+            if node.tags.get("railway") == "tram_stop"
+        ]
+        self._coords_data = tram_stops_and_tracks.nodes
+        self._permanent_nodes = set()
+        self._ways_dict = {}
+
+        self._initialize_graph()
```
RCRalph, line 32: I'm starting to think that we don't actually need this method and we could move it's content right into the constructor:
```suggestion
        self._tram_track_graph = self._build_initial_graph()
        self._permanent_nodes = self._find_permanent_nodes()

        self._remove_unnecessary_nodes()
        self._add_two_way_edges()
        self._add_coordinates_from_data()
        self._tag_nodes_with_way_ids()
        self._build_dict()

        while self._is_there_any_ways_to_merge():
            self._merge_ways(tolerance=2.0)
```

I think it's explicit enough that we're building the graph here, also we could avoid a lot of scrolling if we wanted to know what `_initialize_graph` actually does. What do you think?

server/src/graph_maker/tram_track_graph_transformer.py
RCRalph: Actually, don't add it because of the comment about moving `_initialize_graph` to constructor

server/src/graph_maker/tram_track_graph_transformer.py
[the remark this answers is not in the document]
olobuszolo: Added explanation to the docstring of `_build_initial_graph`

server/src/graph_maker/tram_track_graph_transformer.py
olobuszolo: You're right. Now it is placed above `subdivide_ways`.

server/src/graph_maker/tram_track_graph_transformer.py
olobuszolo: So I didn't add it

server/src/graph_maker/tram_track_graph_transformer.py
olobuszolo: Good point 👍 - applied

server/src/graph_maker/tram_track_graph_transformer.py
olobuszolo: Yeah, definitely  your proposition name is better, applied

server/src/graph_maker/tram_track_graph_transformer.py
olobuszolo: Yeah, as above new name is better idea - applied.

server/src/graph_maker/tram_track_graph_transformer.py
olobuszolo: Definitely. Now everyone knows what those mysterious `"main_nodes"` are. Removed first sentence but rest of this docstring is still there because it explains why we need them.

server/src/graph_maker/tram_track_graph_transformer.py
olobuszolo: Good point 👍 - applied

server/src/graph_maker/tram_track_graph_transformer.py
olobuszolo: You're right. I've changed name of this graph as you suggested to `_tram_track_graph` and added this to constructor.

server/src/graph_maker/tram_track_graph_transformer.py
olobuszolo: You're right — creating a separate method wasn't necessary. Removed `_initialize_graph` and the code now matches your suggestion.

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,385 @@
...
+
+    def __init__(self, tram_stops_and_tracks: overpy.Result):
+        self._ways = tram_stops_and_tracks.ways
+        self._stops = [
+            node
+            for node in tram_stops_and_tracks.nodes
+            if node.tags.get("railway") == "tram_stop"
+        ]
+        self._coords_data = tram_stops_and_tracks.nodes
+        self._permanent_nodes = set()
```
RCRalph, line 29: There's no need for this line here as we're initializing this attribute later.

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,385 @@
...
+        """
+        These nodes form the basis for the initial _tram_track_graph skeleton.
+        They are essential for filtering out unnecessary nodes before the second stage of _tram_track_graph creation,
+        which involves densifying the network between main nodes.
+        """
+        tram_stops = self._get_tram_stop_node_ids_in_graph()
+        crossings = self._get_track_crossing_node_ids_in_graph()
+        return tram_stops | crossings
+
+    def _remove_unnecessary_nodes(self):
```
RCRalph, line 88: Not sure what kind of nodes are unnecessary, maybe this could be explained in the docstring?

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,385 @@
...
+                for successor in successors:
+                    while (
+                        successor not in self._permanent_nodes
+                        and self._tram_track_graph.out_degree(successor) > 0
+                    ):
+                        to_remove.append(successor)
+                        successor = list(self._tram_track_graph.successors(successor))
+                        if len(successor) > 1:
+                            break
+                        successor = successor[0]
```
RCRalph, line 107: It's generally a bad practice to change types of variables between instructions, could you change the variable assignments so this problem is avoided?

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,385 @@
...
+                    and self._tram_track_graph.out_degree(successor) > 0
+                ):
+                    to_remove.append(successor)
+                    successor = list(self._tram_track_graph.successors(successor))
+                    if len(successor) > 1:
+                        break
+                    successor = successor[0]
+                for i in to_remove:
+                    self._tram_track_graph.remove_node(i)
+                self._tram_track_graph.add_edge(node, successor)
```
RCRalph, line 127: This seems to be the exact copy of the code fragment above, which suggest extracting it to a separate method in order to make it more readable, what do you think?

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,385 @@
...
+
+    def _get_tram_stop_node_ids_in_graph(self):
+        return {
+            stop.id
+            for stop in self._stops
+            if stop.id in self._tram_track_graph.nodes
+            and (
+                self._tram_track_graph.out_degree(stop.id) > 0
+                or self._tram_track_graph.in_degree(stop.id) > 0
+            )
```
RCRalph, line 67: Not sure why this condition is here, could you provide some reasoning behind it?

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,385 @@
...
+        }
+
+    def _get_track_crossing_node_ids_in_graph(self):
+        return {
+            node
+            for node in self._tram_track_graph.nodes
+            if self._tram_track_graph.in_degree(node) > 1
+            or self._tram_track_graph.out_degree(node) > 1
```
RCRalph, line 75: What if we defined a track crossing as a node which has more than 2 distinct nodes which are connected either to or from this node? Something like this would do:
```python
len(set(self._tram_track_graph.predecessors(node)) | set(self._tram_track_graph.successors(node))) > 2
```
This way we could avoid the issue with adding two-way connections to the graph before the process of finding permanent nodes. Furthermore, if we changed this condition from `> 2` to `!= 2`, we could also find the nodes which mark the beginning or the end of the track.

I guess it would be more readable if we either renamed this method to `_get_track_crossing_and_buffer_node_ids` or made a different method for buffers all together. What are your thoughts?

server/src/graph_maker/tram_track_graph_transformer.py
olobuszolo: Yeah, I didn't noticed that. Removed this unnecessary line.

server/src/graph_maker/tram_track_graph_transformer.py
olobuszolo: I changed the name to `_remove_non_permanent_nodes` with simple docstring. What do you think about new name?

server/src/graph_maker/tram_track_graph_transformer.py
olobuszolo: Good point, applied your suggestion.

server/src/graph_maker/tram_track_graph_transformer.py
olobuszolo: Yeah, but now the logic in this method has been changed- there's no unnecessary code duplication anymore.

server/src/graph_maker/tram_track_graph_transformer.py
olobuszolo: Probably it was unnecessary condition which I wanted to check. Removed.

server/src/graph_maker/tram_track_graph_transformer.py
RCRalph: Sounds much better now, thanks :3

server/src/graph_maker/tram_track_graph_transformer.py
olobuszolo: Yeah, this idea is correct. It helps us to avoid adding two-way edges after and we don't need to find start nodes. I've applied this suggestion and renamed this method to `_get_track_crossing_and_buffer_node_ids` as you suggested.

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,393 @@
...
+        self._add_coordinates_from_data()
+        self._tag_nodes_with_way_ids()
+        self._build_dict()
+        while self._is_there_any_ways_to_merge():
+            self._merge_ways(tolerance=2.0)
+
+    def get_initial_graph(self):
+        """
+        Zwraca kopię grafu zbudowanego na podstawie tramwajowych torów i przystanków,
+        przed zagęszczeniem (densyfikacją).
```
RCRalph: Please use English in code and comments, keeps it consistent with the rest of the text used in this project.

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,393 @@
...
+        self._tram_track_graph = self._build_initial_graph()
+        self._permanent_nodes = self._find_permanent_nodes()
+        self._remove_non_permanent_nodes()
+        self._add_coordinates_from_data()
+        self._tag_nodes_with_way_ids()
+        self._build_dict()
+        while self._is_there_any_ways_to_merge():
+            self._merge_ways(tolerance=2.0)
+
+    def get_initial_graph(self):
```
RCRalph: Please change this to a property and move it to the appropriate place. Also initial graph isn't very meaningful to me, could you come up with something that describes this better? Like permanent graph or something.

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,386 @@
...
+
+        for way in self._ways:
+            node_ids = [node.id for node in way.nodes]
+            is_oneway = way.tags.get("oneway") == "yes"
+
+            for i in range(len(node_ids) - 1):
+                from_node = node_ids[i]
+                to_node = node_ids[i + 1]
+
+                graph.add_edge(from_node, to_node, oneway=is_oneway)
```
RCRalph, line 57: I don't like adding the `oneway` property to an edge, it makes it unnecessarily complex. In order to achieve that functionality, given an edge between A and B, you could just check if there exists an edge between B and A.

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,386 @@
...
+                if not is_oneway:
+                    graph.add_edge(to_node, from_node, oneway=False)
+
+        return graph
+
+    def _get_tram_stop_node_ids_in_graph(self):
+        return {
+            stop.id for stop in self._stops if stop.id in self._tram_track_graph.nodes
```
RCRalph, line 65: Is this condition ever false? I would honestly be surprised if it could be false.

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,386 @@
...
+        tram_stops = self._get_tram_stop_node_ids_in_graph()
+        crossings = self._get_track_crossing_and_buffer_node_ids()
+        return tram_stops | crossings
+
+    def _remove_non_permanent_nodes(self):
+        """
+        Removes non-permanent intermediate nodes between permanent nodes.
+        """
+
+        def remove_intermediate_nodes_from(permanent_nodes):
```
RCRalph, line 97: Adding a function here serves no purpose I believe.

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,386 @@
...
+        """
+        These nodes form the basis for the initial _tram_track_graph skeleton.
+        They are essential for filtering out unnecessary nodes before the second stage of _tram_track_graph creation,
+        which involves densifying the network between main nodes.
+        """
+        tram_stops = self._get_tram_stop_node_ids_in_graph()
+        crossings = self._get_track_crossing_and_buffer_node_ids()
+        return tram_stops | crossings
+
+    def _remove_non_permanent_nodes(self):
```
RCRalph, line 92: My suggestion is to instead of removing nodes from the current graph, create a new one which only contains the permanent nodes and connections between them and return it from this method.

My idea for this would be to run a DFS-like algorithm for each permanent node until you find another permanent node, then add a one-way connection between these two and cut that search branch. Due do the specifics of which nodes are permanent, we know that there is exactly one path between two neighboring permanent nodes. This approach also works for two-way connections, because we're basically searching which permanent nodes are reachable from a given node.

server/src/graph_maker/tram_track_graph_transformer.py
olobuszolo: Okay, so now `oneway` tag is delated.

server/src/graph_maker/tram_track_graph_transformer.py
olobuszolo: Yeah, for example when tram track is out of service but tram stops are still in OSM data. I added short docstring to this method with explanation

server/src/graph_maker/tram_track_graph_transformer.py
olobuszolo: You're right. I forgot to delate it. Now it looks like it should.

server/src/graph_maker/tram_track_graph_transformer.py
olobuszolo: Okay, it makes sense. So I've implemented DFS-like algorithm for creating new graph only from `_permanent_nodes`. At the end of creating new graph the old one is changed to the new one. Also I've changed name to `_build_graph_with_permanent_nodes`.

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,375 @@
...
+                    current = next_nodes[0]
+                    path.append(current)
+
+                if current in self._permanent_nodes:
+                    new_graph.add_edge(start_node, current)
+
+                    if self._tram_track_graph.has_edge(current, start_node):
+                        new_graph.add_edge(current, start_node)
+
+        self._tram_track_graph = new_graph
```
RCRalph, line 127: I was thinking that we return this graph and assign it to a new class attribute instead of reassigning the already existing one, maybe the original graph could be useful in the second part of the implementation.

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,375 @@
...
+        """
+        tram_stops = self._get_tram_stop_node_ids_in_graph()
+        crossings = self._get_track_crossing_and_buffer_node_ids()
+        return tram_stops | crossings
+
+    def _build_graph_with_permanent_nodes(self):
+        new_graph = nx.DiGraph()
+
+        for start_node in self._permanent_nodes:
+            for neighbor in self._tram_track_graph.successors(start_node):
```
RCRalph, line 99: I have an idea of a nice refactor of this method but it's quite complicated to explain exactly what I mean, can I make changes to this fragment and then you'll review it?

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,375 @@
...
+
+                if current in self._permanent_nodes:
+                    new_graph.add_edge(start_node, current)
+
+                    if self._tram_track_graph.has_edge(current, start_node):
+                        new_graph.add_edge(current, start_node)
+
+        self._tram_track_graph = new_graph
+
+    def _is_there_any_ways_to_merge(self):
```
RCRalph, line 129: I think this method should come a bit later in the class as it's using things not defined just yet

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,375 @@
...
+            if count < 2:
+                return True
+        return False
+
+    def _add_coordinates_from_data(self):
+        nodes_dict = {
+            int(node.id): (float(node.lat), float(node.lon))
+            for node in self._coords_data
+            if node.lat is not None and node.lon is not None
+        }
```
RCRalph, line 145: From what I can see we're using the `self._coords_data` property only for this kind of transformation. Maybe it's a good idea to make a protected attribute `_node_coordinates_by_id` initialized in the constructor which would contain this exact data? Then this method would have very little use, since we would be able to get all the information about node coordinates using `self._node_coordinates_by_id[node_id]`.

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,375 @@
...
+            way_id = way.id
+            for node in way.nodes:
+                node_id = node.id
+                if node_id in self._tram_track_graph.nodes:
+
+                    if "ways" not in self._tram_track_graph.nodes[node_id]:
+                        self._tram_track_graph.nodes[node_id]["ways"] = []
+                    self._tram_track_graph.nodes[node_id]["ways"].append(way_id)
+
+    def _build_dict(self):
```
RCRalph, line 165: What kind of dict are we building here? I think it should be described in the docstring.

server/src/graph_maker/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,375 @@
...
+        self._ways_dict = {}
```
RCRalph, line 166: If we're building `self._ways_dict` (whatever it is) from scratch, then maybe we could just return it from the method and perform regular attribute assignment in the class initializer?

server/src/graph_maker/tram_track_graph_transformer.py
olobuszolo: Sure

RCRalph: Added a commit which refactors this method

server/src/graph_maker/tram_track_graph_transformer.py
RCRalph: Assigned it to `self._minified_tram_track_graph`

server/src/graph_maker/tram_track_graph_transformer.py
olobuszolo: Yeah, it' s good idea with making a protected attribute `_node_coordinates_by_id`. Added this atribute

server/src/graph_maker/tram_track_graph_transformer.py
Redor144: Good point. Moving it down makes sense.

server/src/graph_maker/tram_track_graph_transformer.py
Redor144: A simple docstring explaining the dict would help.

server/src/graph_maker/tram_track_graph_transformer.py
Redor144: Yeah, that’s cleaner. Assigning self._ways_dict = self._build_ways_dict() makes it clear where it comes from.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,382 @@
...
+        return self._permanent_nodes.copy()
+
+    def _subdivide_ways(self, _tram_track_graph: nx.DiGraph, max_distance: float):
+        edges_to_remove = set()
+
+        def find_existing_node_by_coords(lat, lon, tolerance=1e-6):
+            for node_id, data in _tram_track_graph.nodes(data=True):
+                if (
+                    abs(data.get("lat", 0) - lat) < tolerance
+                    and abs(data.get("lon", 0) - lon) < tolerance
```
RCRalph, line 288: How about assigning `abs(data.get("lat", 0) - lat)` and `abs(data.get("lon", 0) - lon)` to separate variables? I think it might improve on readability a bit.

Also let's maybe use a dict similar to `_node_coordinates_by_id` as the approach with storing data in graph nodes is problematic from the perspective of type hints.

You could also create a custom class in the `model` package which will contain the node ID, latitude and longitude. The nodes in the graph returned from your method would then contain nodes which would be instances of your class. In this data structure you could then define a method called `is_close_to`, where you would compare the coordinates with a default tolerance of `1e-6`.

What do you think?

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,382 @@
...
+    def __init__(self, tram_stops_and_tracks: overpy.Result):
+        self._ways = tram_stops_and_tracks.get_ways()
+        self._stops = [
+            node
+            for node in tram_stops_and_tracks.get_nodes()
+            if node.tags.get("railway") == "tram_stop"
+        ]
+        self._coords_data = tram_stops_and_tracks.get_nodes()
+        self._node_coordinates_by_id = {
+            int(node.id): (float(node.lat), float(node.lon))
```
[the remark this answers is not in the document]
olobuszolo: You're right. Removed unnecessary converting.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,382 @@
...
+                continue
+
+            linestring = LineString(coords)
+
+            if way.id in ways_dict:
+                existing = ways_dict[way.id]
+                ways_dict[way.id] = linemerge([existing, linestring])
+            else:
+                ways_dict[way.id] = linestring
+        return ways_dict
```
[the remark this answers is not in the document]
Redor144: I have change the way how we densify the graph and now adding new nodes with `max_distance_in_meters = 25` take something around 6-7 seconds. Now method `densify_graph_by_max_distance` add new nodes only on one way tracks and I am working on to handle bi-driectional tracks.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,382 @@
...
+        edges_to_remove = set()
+
+        def find_existing_node_by_coords(lat, lon, tolerance=1e-6):
+            for node_id, data in _tram_track_graph.nodes(data=True):
+                if (
+                    abs(data.get("lat", 0) - lat) < tolerance
+                    and abs(data.get("lon", 0) - lon) < tolerance
+                ):
+                    return node_id
+            return None
```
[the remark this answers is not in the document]
Redor144: In new approach we do not use something like this.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
Redor144: As in previous comments the change of approach eliminated this problem.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,241 @@
...
+
+            interpolated_points = [line.interpolate(d) for d in distances]
+            interpolated_latlon = [
+                inverse.transform(p.x, p.y)[::-1] for p in interpolated_points
+            ]
+
+            return [latlon[0]] + interpolated_latlon + [latlon[-1]]
+
+        new_graph = nx.DiGraph()
+        new_id_counter = max(self._tram_track_graph.nodes) + 1
```
RCRalph, line 188: I think we could use the `self._permanent_nodes` property here, it should provide better typing information:
```suggestion
        new_id_counter = max(self._permanent_nodes) + 1
```

server/requirements.txt
```diff
@@ -2,3 +2,6 @@ overpy>=0.7
 pandas>=2.2.3
 pydantic>=2.10.6
 requests>=2.32.3
+networkx>=3.4.2
+pyproj >= 3.7.1
+shapely >= 2.0.7
```
RCRalph, line 7: Please also remove the whitespaces here, they conflict with the rest of the file
```suggestion
pyproj>=3.7.1
shapely>=2.0.7
```

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,241 @@
...
+
+            num_segments = max(1, math.ceil(length / max_distance_in_meters))
+            distances = [i * (length / num_segments) for i in range(1, num_segments)]
+
+            interpolated_points = [line.interpolate(d) for d in distances]
+            interpolated_latlon = [
+                inverse.transform(p.x, p.y)[::-1] for p in interpolated_points
+            ]
+
+            return [latlon[0]] + interpolated_latlon + [latlon[-1]]
```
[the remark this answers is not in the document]
Redor144: Sure, applying it.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,241 @@
...
+
+                while node not in self._permanent_nodes:
+                    path.append(
+                        (
+                            node,
+                            self._node_coordinates_by_id[node][0],
+                            self._node_coordinates_by_id[node][1],
+                        )
+                    )
+                    node = list(self._tram_track_graph.successors(node))[0]
```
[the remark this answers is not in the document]
Redor144: Now it will also look nice  thought.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,241 @@
...
+                    lat, lon = interpolated_nodes[i]
+                    if i == 0:
+                        new_graph.add_node(u, lat=lat, lon=lon, permanent=True)
+                        prev_node = u
+                    elif i == len(interpolated_nodes) - 1:
+                        new_graph.add_node(node, lat=lat, lon=lon, permanent=True)
+                        new_graph.add_edge(prev_node, node)
+                    else:
+                        new_graph.add_node(
+                            new_id_counter, lat=lat, lon=lon, permanent=False
```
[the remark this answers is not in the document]
Redor144: Exactly it was used for debugging, just forgot about it.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,241 @@
...
+        self._coords_data = tram_stops_and_tracks.get_nodes()
+        self._node_coordinates_by_id = {
+            node.id: (float(node.lat), float(node.lon))
+            for node in self._coords_data
+            if node.lat is not None and node.lon is not None
+        }
+
+        self._tram_track_graph = self._build_tram_track_graph_from_osm_ways()
+        self._permanent_nodes = self._find_permanent_nodes()
+        self._minified_tram_track_graph = self._build_minified_tram_track_graph()
```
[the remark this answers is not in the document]
Redor144: Maybe `self._minified_tram_track_graph` is now not needed, but let it stay until the bi-directional tracks will be working. I don't thing it will be useful with that, but keep it just for a while. What you thing?

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,241 @@
...
+                    )
+
+    @property
+    def permanent_node_ids(self):
+        return self._permanent_nodes.copy()
+
+    def densify_graph_by_max_distance(
+        self, max_distance_in_meters: float
+    ) -> nx.DiGraph:
+
```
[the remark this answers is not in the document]
Redor144: You're right, this method should contains docstring.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
Redor144: Good point. Changed.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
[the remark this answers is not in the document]
RCRalph: Ah so the bidirectional tracks still aren't working, in this case let's keep it until we have the final implementation

server/requirements.txt
Redor144: 👍

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,241 @@
...
+                        self._node_coordinates_by_id[node][1],
+                    )
+                )
+                interpolated_nodes = interpolate_between_first_and_last_node(
+                    path, max_distance_in_meters
+                )
+
+                for i in range(len(interpolated_nodes)):
+                    lat, lon = interpolated_nodes[i]
+                    if i == 0:
```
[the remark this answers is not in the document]
Redor144: I don't see that, it would consist `case _ if i == len(interpolated_nodes) - 1: `. Imo that will impair readability. Maybe I don't get something.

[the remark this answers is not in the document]
RCRalph: How about:
```py
match i:
    case 0:
        ...
    case len(interpolated_nodes) - 1:
        ...
    case _:
        ...
```

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,241 @@
...
+
+                if self._tram_track_graph.has_edge(v, u):
+                    continue
+
+                path = [
+                    (
+                        u,
+                        self._node_coordinates_by_id[u][0],
+                        self._node_coordinates_by_id[u][1],
+                    )
```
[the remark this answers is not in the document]
Redor144: You mean something like this?
```suggestion
                    Node(
                        node_id=u,
                        lat=self._node_coordinates_by_id[u][0],
                        lon=self._node_coordinates_by_id[u][1],
                    )
```
Then only `_interpolate_between_first_and_last_node` method have to be changed.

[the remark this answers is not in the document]
Redor144: Even simpler would be to add `self._node_coordinates_by_id[u]` to path because it is tuple with lat and lon.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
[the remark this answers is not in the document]
Redor144: With new approach which handle bidirectional tracks `self._minified_tram_track_graph` and `self._build_minified_tram_track_graph` are unnecessary. Deleted

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,217 @@
...
+
+        Args:
+            max_distance_in_meters (float):
+                The maximum allowed distance between consecutive points after interpolation.
+
+        Returns:
+            networkx.DiGraph:
+                The newly built, interpolated directed graph.
+        """
+        new_graph = nx.DiGraph()
```
RCRalph, line 151: I believe in your imagination that you can come up with a better name for this variable :3

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,217 @@
...
+                        if next_candidate != previous_node:
+                            previous_node = current_node
+                            current_node = next_candidate
+                            break
+                    else:
+                        path_found = False
+                        break
+
+                if not path_found:
+                    continue
```
RCRalph, line 179: Is this condition ever true? I need to see an example of a path not being found because to me it sounds like an exception.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,217 @@
...
+
+                current_node = successor
+                previous_node = permanent_node
+                path_found = True
+
+                while current_node not in self._permanent_nodes:
+                    path_coordinates.append(self._node_coordinates_by_id[current_node])
+                    successors_list = list(
+                        self._tram_track_graph.successors(current_node)
+                    )
```
RCRalph, line 167: No need to create a list as the we're then directly iterating through it.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,217 @@
...
+                        else:
+                            last_graph_node = current_node
+                            new_graph.add_node(last_graph_node, lat=lat, lon=lon)
+
+                        new_graph.add_edge(previous_graph_node, last_graph_node)
+
+                    else:
+                        if (lat, lon) in coord_map:
+                            new_node = coord_map[(lat, lon)]
+                            new_graph.add_node(new_node, lat=lat, lon=lon)
```
RCRalph, line 208: I would still like to see the `Node` datastructure here if possible.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,217 @@
...
+            max_distance_in_meters (float):
+                The maximum allowed distance between consecutive points after interpolation.
+
+        Returns:
+            networkx.DiGraph:
+                The newly built, interpolated directed graph.
+        """
+        new_graph = nx.DiGraph()
+        new_node_id_counter = max(self._permanent_nodes) + 1
+        coord_map = {}
```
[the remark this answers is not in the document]
Redor144: 👍

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,217 @@
...
+        Builds a new directed graph where each segment between permanent nodes is split
+        into smaller segments according to the specified distance limit.
+
+        Args:
+            max_distance_in_meters (float):
+                The maximum allowed distance between consecutive points after interpolation.
+
+        Returns:
+            networkx.DiGraph:
+                The newly built, interpolated directed graph.
```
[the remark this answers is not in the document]
Redor144: I'm taking care of it.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,217 @@
...
+        distances = [i * (length / num_segments) for i in range(1, num_segments)]
+
+        interpolated_points = [line.interpolate(d) for d in distances]
+        interpolated_latlon = [
+            inverse.transform(p.x, p.y)[::-1] for p in interpolated_points
+        ]
+
+        return [latlon[0]] + interpolated_latlon + [latlon[-1]]
+
+    def _densify_graph_by_max_distance(self, max_distance_in_meters):
```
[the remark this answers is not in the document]
Redor144: Adding missing things.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,217 @@
...
+            ):
+                stack.append(item)
+
+        return result
+
+    @property
+    def permanent_node_ids(self):
+        return self._permanent_nodes.copy()
+
+    def _interpolate_between_first_and_last_node(self, latlon, max_distance_in_meters):
```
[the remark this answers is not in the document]
Redor144: Of course

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,217 @@
...
+            if node in self._permanent_nodes:
+                result.add(node)
+                continue
+
+            for item in filter(
+                lambda x: x not in visited, self._tram_track_graph.successors(node)
+            ):
+                stack.append(item)
+
+        return result
```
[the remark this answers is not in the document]
Redor144: Exactly, removed.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,247 @@
...
+        ]
+
+        return [node_coordinates[0]] + interpolated_latlon + [node_coordinates[-1]]
+
+    def _add_interpolated_nodes_with_edges(
+        self,
+        densified_graph: nx.DiGraph,
+        interpolated_nodes: list[tuple[float, float]],
+        coord_map: dict[tuple[float, float], Node],
+        new_node_id_counter: int,
```
RCRalph, line 160: I think there is a strong case here for extracting this variable to a class property and adding a protected metod `_get_new_node_id`, which would increment the variable and return the value.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,247 @@
...
+from pyproj import Transformer
+from shapely.geometry import LineString
+
+from .exceptions import TrackDirectionChangeError
+from .model import Node
+
+
+class TramTrackGraphTransformer:
+    """
+    TramTrackGraphTransformer processes OpenStreetMap (OSM) tram infrastructure data
```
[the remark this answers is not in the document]
Redor144: Yeah, that's so true.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,247 @@
...
+        """
+
+        tram_stops = self._get_tram_stop_node_ids_in_graph()
+        crossings = self._get_track_crossing_and_endpoint_node_ids()
+
+        return tram_stops | crossings
+
+    def _find_path_between_permenent_nodes(
+        self, permanent_node: Node, successor: Node
+    ) -> tuple[list[int], Node]:
```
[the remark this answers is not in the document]
Redor144: Changed to `tuple[list[tuple[float, float]], Node]`, but now it looks illegible. Maybe remove it, what do you think?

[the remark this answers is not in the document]
RCRalph: I suggest switching to `list[Node]` as per my other comment

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,247 @@
...
+                    new_node_id_counter = self._add_interpolated_nodes_with_edges(
+                        densified_graph,
+                        interpolated_nodes,
+                        coord_map,
+                        new_node_id_counter,
+                        permanent_node,
+                        last_node,
+                    )
+                except TrackDirectionChangeError as e:
+                    errors.append(e)
```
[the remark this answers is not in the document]
Redor144: Ok, so it would look like this:
```python
try:
    path_coordinates = (
        self._find_path_between_permenent_nodes(
            permanent_node, successor
        )
    )
except TrackDirectionChangeError as e:
    errors.append(e)
    continue
...
```
Because, we can't add new nodes without correct path.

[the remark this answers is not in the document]
RCRalph: That's what I had in mind

server/src/tram_track_graph_transformer/model.py
```diff
@@ -0,0 +1,8 @@
+from pydantic import BaseModel
+
+
+class Node(BaseModel, frozen=True):
+    id: int
+    lat: float
+    lon: float
+    type: str
```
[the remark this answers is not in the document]
Redor144: Good point, create another file for enums or place it in `node.py`?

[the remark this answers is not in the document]
RCRalph: Another file would be nice, it's not like a file costs 5 bucks.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,247 @@
...
+    """
+
+    def __init__(self, tram_stops_and_tracks: overpy.Result):
+        self._ways = tram_stops_and_tracks.get_ways()
+        self._nodes_by_id = {
+            node.id: Node(
+                id=node.id,
+                lat=float(node.lat),
+                lon=float(node.lon),
+                type=node.tags.get("railway", ""),
```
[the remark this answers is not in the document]
Redor144: Default value will be `NodeType.UNKNOWN ` where `NodeType` is enum.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,247 @@
...
+            - type (str): type of the node, e.g., "tram_stop", "switch", etc.
+
+        Edges: no attributes.
+        """
+        if max_distance_in_meters <= 0:
+            raise ValueError("max_distance_in_meters must be greater than 0.")
+
+        densified_graph: nx.DiGraph = nx.DiGraph()
+        new_node_id_counter: int = max(node.id for node in self._permanent_nodes) + 1
+        coord_map: dict[tuple[float, float], Node] = {}
```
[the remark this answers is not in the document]
Redor144: It's used in adding bidirectional tracks. When we processed edge A -> B and now we are add nodes B -> A on the same track `coord_map` ensure that we don't create two nodes in the same coordinates.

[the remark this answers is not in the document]
RCRalph: The comment is probably in the wrong place, I'm OK with keeping this dict as it is now but please change the name of this variable.

server/src/tram_track_graph_transformer/exceptions.py
```diff
@@ -0,0 +1,9 @@
+class TrackDirectionChangeError(Exception):
+    """
+    Exception raised when a change in track direction is detected
+    at a node where such a change is not allowed.
+    """
+
+    def __init__(self, node_id: int):
```
RCRalph, line 7: Would be nice to know from which direction the change is originating (what is the source permanent node)

RCRalph (approved):
We got there in the end :3

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -0,0 +1,247 @@
...
+
+                graph.add_edge(source_node, destination_node)
+                if not is_oneway:
+                    graph.add_edge(destination_node, source_node)
+
+        return graph
+
+    def _get_tram_stop_node_ids_in_graph(self):
+        """
+        Returns set of tram stop node which are on the tram tracks provided by OSM.
```
olobuszolo, line 66:
```suggestion
        Returns set of tram stop nodes which are on the tram tracks provided by OSM.
```

olobuszolo (approved):
LGTM ;]

Issue #18, opened 2025-04-02
Implement tram stop reachability check for all known tram trips


RCRalph:
### Description
In order to be sure that our data processing pipeline produced valid simulation data, we have to be sure that every proposed tram trip can be completed. This can be checked by trying to find any path between every pair of stops which appear in sequence. In case a path like this doesn't exist, an exception should be raised. We should also assume some maximal distance between two stops in order to avoid long loops.

The information about tram stops of trips can be found by using the `TramStopMapper.get_stop_nodes_by_gtfs_trip_id` method. The tram track graph should come from the `TramTrackGraphTransformer.densify_graph_by_max_distance` method.

Tests for this functionality should also be created, both for the passing and failing cases.

### Definition of done
We have a function / class, which we can use to verify if it's possible to conduct a simulation with provided data. This check should complete successfully only if it's possible for a tram to complete the given trip. This functionality has unit test coverage.

### Notes
You can use any graph traversal algorithm, but A* with the heuristic of euclidean coordinates seems to be the most optimal.

PR #24, opened 2025-04-07
Olobuszolo/tram stop reachability tests

olobuszolo:
### Related issues
- #18

### Short description
Added tram reachability tests, covering cases with no path between nodes, excessively long paths, and valid paths for all known pairs.


server/tests/tram_stop_reachability/test_tram_stop_reachability.py
```diff
@@ -0,0 +1,144 @@
...
+        lons = [node.lon for node in path]
+        lats = [node.lat for node in path]
+        return self.geod.line_length(lons, lats)
+
+    def test_graph_reachability(self):
+        """
+        Test if all pairs of tram stops are reachable within the allowed distance.
+        Special pairs are pairs which are close but path goes through a longer route e.g Suche Stawy 01 to Bardosa 02.
+        Those pairs have a special ratio which is higher than the default one.
+        """
```
RCRalph, line 89: These doc strings aren't necessary I think, what is being tested should be known from the test method name itself

server/tests/tram_stop_reachability/test_tram_stop_reachability.py
RCRalph, line 1: The goal of this PR should be to introduce a function/class/method into the *production code* instead of the test themselves. All I can see in this PR are tests, while this functionality should be a part of the data processing pipeline. My suggestion is to add a static/class method to `TramTrackGraphTransformer` class called `verify_node_reachability`, which would check the presented conditions.

RCRalph (changes requested):

server/tests/tram_stop_reachability/test_tram_stop_reachability.py
RCRalph: Or maybe create a new class in that same package, your call.

RCRalph: Or maybe just have a function which adds that functionality, your call as well.

server/tests/tram_stop_reachability/test_tram_stop_reachability.py
olobuszolo: That's right. Removed doc string

server/tests/tram_stop_reachability/test_tram_stop_reachability.py
olobuszolo: So I've created method `check_path_viability` in `TramTrackGraphTransformer` which check if there exist path between two nodes.

server/src/model/city_configuration.py
```diff
@@ -1,8 +1,18 @@
-from pydantic import BaseModel
+from typing import List
+
+from pydantic import BaseModel, Field
+
+
+class TramStopPairCheck(BaseModel):
+    from_: int = Field(alias="from")
```
RCRalph, line 7: Please add a comment that `from` is a reserved keyword in Python since it's not obvious at the first sight.

Also in this case maybe it's a good idea to rename these fields to source and destination, your call.

server/src/model/city_configuration.py
```diff
@@ -1,8 +1,18 @@
...
+    ratio: float
 
 
 class CityConfiguration(BaseModel):
     osm_area_name: str
     gtfs_url: str
     ignored_gtfs_lines: list[str]
     custom_stop_mapping: dict[str, int]
+    max_distance_ratio: float
+    custom_tram_stop_pair_max_distance_checks: List[TramStopPairCheck]
```
RCRalph, line 18: Importing `List` from `typing` is deprecated since Python 3.9 I believe:
```suggestion
    custom_tram_stop_pair_max_distance_checks: list[TramStopPairCheck]
```

server/src/tram_track_graph_transformer/exceptions.py
```diff
@@ -11,3 +11,21 @@ def __init__(self, start_node_id: int, node_id: int):
             f"Track from permanent node {start_node_id} "
             f"changes direction at non-permanent node {node_id}."
         )
+
+
+class NoPathFoundError(Exception):
+    def __init__(self, start, end):
```
RCRalph, line 17: Please add type hints here

server/src/tram_track_graph_transformer/exceptions.py
```diff
@@ -11,3 +11,21 @@ def __init__(self, start_node_id: int, node_id: int):
...
+        super().__init__(f"No path found between stops: {start} -> {end}")
+
+
+class PathTooLongError(Exception):
+    def __init__(self, start, end, actual, allowed):
```
RCRalph, line 22: Missing type hints here also

server/src/tram_track_graph_transformer/exceptions.py
```diff
@@ -11,3 +11,21 @@ def __init__(self, start_node_id: int, node_id: int):
...
+        super().__init__(
+            f"Path too long: {start} -> {end}\r\n"
```
RCRalph, line 24: I think we shouldn't be using `\r\n` anywhere since the server will be running on a Linux system, which uses `\n` as newline character.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -61,6 +71,14 @@ def _build_tram_track_graph_from_osm_ways(self):
 
         return graph
 
+    def _get_custom_stop(self):
+        return list(self._city_configuration.custom_stop_mapping.values())
```
RCRalph, line 75: This looks more like a property of the `CityConfiguration` class rather than a method of this class. Also I believe we could also skip calling an external method for this and just using the `.values()` directly, what do you think?

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -245,3 +263,47 @@ def densify_graph_by_max_distance(
...
+        )
+        return path
+
+    @classmethod
+    def check_path_viability(
+        cls, graph: nx.DiGraph, start_id: int, end_id: int, ratio: float
+    ) -> bool:
+        geod = Geod(ellps="WGS84")
+        start_node = next((n for n in graph.nodes if n.id == start_id), None)
+        end_node = next((n for n in graph.nodes if n.id == end_id), None)
```
RCRalph, line 287: This is very costly in terms of performance as this technically becomes a `O(n^2)` algorithm. The easiest way to avoid this would be to use a `nodes_by_id` dict.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -245,3 +263,47 @@ def densify_graph_by_max_distance(
...
+
+        for node_id, node in [(start_id, start_node), (end_id, end_node)]:
+            if node is None:
+                raise NodeNotFoundError(node_id)
```
RCRalph, line 291: I think this loop is a bit overkill, I suggest doing something like this:
```py
start_node = ...
if start_node is None:
    raise NodeNotFoundError(start_node_id)

end_node = ...
if end_node is None:
    raise NodeNotFoundError(end_node_id)
```

You can also try using the Walrus operator (`:=`) here, but not sure if it will look better.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -245,3 +263,47 @@ def densify_graph_by_max_distance(
...
+        )[2]
+
+        lons = [node.lon for node in path]
+        lats = [node.lat for node in path]
+        path_length = geod.line_length(lons, lats)
+        if path_length > straight_line_distance * ratio:
+            raise PathTooLongError(
+                start_id, end_id, path_length, straight_line_distance * ratio
+            )
+        return True
```
RCRalph, line 309: If we're only returning `True` or raising an exception, I believe there's no need to return anything as a successful execution will not raise an exception and an unsuccessful one will raise one.

server/tests/tram_stop_reachability/test_tram_stop_reachability.py
RCRalph: As I'm seeing the implementation now, it looks to me like a separate class `TramTrackGraphInspector` 
(I really like that name :3) would be a better option for this functionality, it could be in the same package as `TramTrackGraphTransformer`.

server/src/model/city_configuration.py
olobuszolo: Renamed `from_` to `source` and `to` to `destination`

server/src/model/city_configuration.py
olobuszolo: Changed to `list`

server/src/tram_track_graph_transformer/exceptions.py
olobuszolo: Added type hint

server/src/tram_track_graph_transformer/exceptions.py
olobuszolo: Okay, so I've removed this and now entire string is on one line.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
olobuszolo: Sure, so removed this methos and now we using the `values()` directly in `_get_node_type`.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
olobuszolo: Yeah, Walrus looks much better, applied.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
olobuszolo: Yeah, removed returning True.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -61,6 +67,11 @@ def _build_tram_track_graph_from_osm_ways(self):
 
         return graph
 
+    def _get_node_type(self, node: Node):
```
RCRalph, line 70: Add a short docstring here:

```suggestion
    def _get_node_type(self, node: Node):
        """
        OSM node IDs mentioned in `custom_stop_mapping` field of `CityConfiguration`
        should always have a type of `NodeType.TRAM_STOP`. This is due to these nodes
        being used for tram stop mapping between GTFS and OSM, which means that
        each node can potentially act as a tram stop, even when it has a different type
        assigned to it by OSM.
        """
```

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -61,6 +67,11 @@ def _build_tram_track_graph_from_osm_ways(self):
...
+        if node.id in self._city_configuration.custom_stop_mapping.values():
+            return NodeType.TRAM_STOP
+        return NodeType.get_by_value_safe(node.tags.get("railway"))
```
RCRalph, line 73: A ternary operator wouldn't look to bad here:
```suggestion
        return (
            NodeType.TRAM_STOP
            if node.id in self._city_configuration.custom_stop_mapping.values()
            else NodeType.get_by_value_safe(node.tags.get("railway"))
        )
```

RCRalph (commented):
Probably last comments about the implementation, I'll try to look into the tests in the evening :3

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -4,6 +4,7 @@
 import overpy
 from pyproj import Transformer
 from shapely.geometry import LineString
+from src.model.city_configuration import CityConfiguration
```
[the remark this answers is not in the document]
olobuszolo: I've overlooked this comment but applied this suggestion, also with other imports

server/src/tram_track_graph_transformer/tram_track_graph_inspector.py
```diff
@@ -0,0 +1,56 @@
...
+        path = nx.astar_path(
+            self.graph,
+            start_node,
+            end_node,
+            heuristic=lambda u, v: self.geod.inv(u.lon, u.lat, v.lon, v.lat)[2],
+            weight=lambda u, v, _: self.geod.inv(u.lon, u.lat, v.lon, v.lat)[2],
+        )
+        return path
+
+    def check_path_viability(self, start_id: int, end_id: int, ratio: float):
```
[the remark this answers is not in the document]
olobuszolo: Changed name to `start_stop_id`, `end_stop_id`, `max_length_ratio`

server/src/tram_track_graph_transformer/tram_track_graph_inspector.py
```diff
@@ -0,0 +1,56 @@
...
+    NoPathFoundError,
+    PathTooLongError,
+)
+from src.tram_track_graph_transformer.node import Node
+
+
+class TramTrackGraphInspector:
+    def __init__(self, graph: nx.DiGraph):
+        self.graph = graph
+        self.geod = Geod(ellps="WGS84")
```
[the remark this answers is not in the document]
olobuszolo: Applied

server/tests/tram_stop_reachability/test_tram_stop_reachability.py
```diff
@@ -0,0 +1,153 @@
...
+        "PlacWszystkichSwietych01_Filharmonia01.zip",
+    ]
+    FILES_WITH_TOO_LONG_PATHS = [
+        "Lubicz02_UniwersytetEkonomiczny01.zip",
+        "TeatrBagatela03_StaryKleparz02.zip",
+    ]
+    FILES_WITH_NO_NODE = [
+        "UKEN_02.zip",
+        "TeatrVariete_02.zip",
+    ]
```
RCRalph, line 41: These test cases you're created look nice, but in my opinion we should think about adding test cases which modify the graph during test runtime instead of using predefined data. This way it should be clearer as of what's actually happening and what data do these tests actually contain. By that I mean let's imagine a test that looks like this:
- Get the tram rail network graph from a fixture
- Remove a critical edge from the graph
- Transform the data
- Expect an exception from the inspector

I will also make these changes to the tram stop mapping tests as we've changed the approach for frozen data, but that's maybe after this PR is merged.

server/tests/tram_stop_reachability/test_tram_stop_reachability.py
```diff
@@ -0,0 +1,153 @@
...
+    def _dijkstra_path(self, graph, start, end):
+        geod = Geod(ellps="WGS84")
+        return nx.dijkstra_path(
+            graph,
+            start,
+            end,
+            weight=lambda u, v, _: geod.inv(u.lon, u.lat, v.lon, v.lat)[2],
+        )
+
+    def test_graph_reachability(self):
```
RCRalph, line 105: Test method names should generally follow a convention:
`test_{method_name}_{short_description(used for fails for example)}`

So this test should be called `test_check_path_viability`.

server/tests/tram_stop_reachability/test_tram_stop_reachability.py
```diff
@@ -0,0 +1,153 @@
...
+
+        for start_id, end_id in pairs:
+            ratio = custom_nodes_dict.get((start_id, end_id), max_ratio)
+            tram_graph_insector = TramTrackGraphInspector(graph)
+            tram_graph_insector.check_path_viability(start_id, end_id, ratio)
+
+    @pytest.mark.parametrize(
+        "file_name", FILES_WITH_NO_NODE + FILES_WITH_NO_PATH + FILES_WITH_TOO_LONG_PATHS
+    )
+    def test_graph_unreachability(self, file_name: str):
```
RCRalph, line 124: Following above comment, this test should be called `test_check_path_viability_node_not_found` for example. It's also a good idea to split this test so it expects just one exception.

server/tests/tram_stop_reachability/test_tram_stop_reachability.py
olobuszolo: Applied your suggestion. Now they are more "flexible"

server/tests/tram_stop_reachability/test_tram_stop_reachability.py
olobuszolo: Good point, applied

server/tests/tram_stop_reachability/test_tram_stop_reachability.py
olobuszolo: Split this method into separate exceptions and renamed them as you suggested

server/tests/tram_stop_reachability/test_tram_stop_reachability.py
```diff
@@ -0,0 +1,194 @@
...
+            == f"No path found between stops: {start_stop} -> {end_stop}"
+        )
+
+    def test_shortest_path_between_nodes(
+        self,
+        krakow_tram_network_graph: nx.DiGraph,
+        tram_stop_pairs: set[tuple[int, int]],
+    ):
+        # Arrange
+        tram_graph_insector = TramTrackGraphInspector(krakow_tram_network_graph)
```
RCRalph, line 180: Please check if this typo isn't present in other places 
```suggestion
        tram_graph_inspector = TramTrackGraphInspector(krakow_tram_network_graph)
```

server/tests/tram_stop_reachability/test_tram_stop_reachability.py
```diff
@@ -0,0 +1,194 @@
...
+        # Act
```
RCRalph, line 181: The Act part of this test only applies to the actual call to the method, the rest is arrange

server/tests/tram_stop_reachability/test_tram_stop_reachability.py
```diff
@@ -0,0 +1,194 @@
...
+        nodes_by_id = {node.id: node for node in krakow_tram_network_graph.nodes}
+        for start_id, end_id in tram_stop_pairs:
+            start_node = nodes_by_id[start_id]
+            end_node = nodes_by_id[end_id]
```
RCRalph, line 185: I think these variables are unnecessary, let's maybe move the values directly to method call, what do you think?

server/tests/tram_stop_reachability/test_tram_stop_reachability.py
```diff
@@ -0,0 +1,194 @@
...
+        ((6738229788, 6738229790), 2419959769, 2420069703),
+        ((2420200294, 8551858404), 2420200261, 2420233975),
+        ((2419367476, 8551888613), 11271425380, 2419894822),
+    ]
+
+    NO_PATH_FOUND_CASES = [
+        ((12685345137, 12685345138), 1768224703, 1768224656),
+        ((12685345293, 12685345294), 2420286331, 2423789750),
+        ((12685341641, 12685341642), 2419986545, 2419986538),
+    ]
```
RCRalph, line 27: You can use `pytest.param` to add meaningful names to your parametrization cases. These names should use snake_case.

server/tests/assets/tram_trips_by_id.json
RCRalph, line 1: Maybe it will be a good idea to zip this file and unzip it in the fixture itself, this way we'll save on space, what do you think? Just make sure you use level 9 compression (default in Python)

server/tests/tram_stop_reachability/test_tram_stop_reachability.py
```diff
@@ -0,0 +1,194 @@
...
+        )
+
+    def test_check_path_viability(
+        self,
+        krakow_city_configuration: CityConfiguration,
+        krakow_tram_network_graph: nx.DiGraph,
+        tram_stop_pairs: set[tuple[int, int]],
+    ):
+        # Arrange
+        custom_nodes_dict = {
```
RCRalph, line 53: As you're using this dict in many places here, maybe it's a good idea to add a property like this to CityConfiguration class? Just make sure to give it a more descriptive name.

olobuszolo: Yeah, added property to `CityConfiguration` named `custom_tram_stop_pair_ratio_map`

server/tests/assets/tram_trips_by_id.json
olobuszolo: Removed the JSON file, added the zipped version, and updated the fixture accordingly.

server/tests/tram_stop_reachability/test_tram_stop_reachability.py
olobuszolo: Ok, now we're using `pytest.param` with meaningful names

server/tests/tram_stop_reachability/test_tram_stop_reachability.py
olobuszolo: Yeah, good point, applied

server/tests/tram_stop_reachability/test_tram_stop_reachability.py
olobuszolo: Replaced Act part just above calling `shortest_path_between_nodes`

server/tests/tram_stop_reachability/test_tram_stop_reachability.py
olobuszolo: Fixed the typo everywhere (I think)

RCRalph (approved):

Redor144 (approved):

Issue #11, opened 2025-03-06
Create tests for tram rail network transformations


RCRalph:
### Description
Following #8, we have to make sure that the proposed implementation of graph transformation is meeting all the requirements. In order to increase our confidence in this process, we should create tests for that feature. The tests should focus on Kraków's tram network, as this is the main area of our interests, but we may also test networks in other cities for extra coverage.

The tests should cover:
- If all tram stops from OSM are present in the graph, except cases where a tram stop is on a disused track or one under construction
- If all intersections from OSM are present int the graph, except cases where an intersection is on/with a disused track or one under construction
- If the densification process yields correctly spaced out nodes
  - The spacing cannot be greater than a provided value
  - The spacing should be even, that means between each two permanent nodes the standard deviation should be minimal

The test should be written in such a way so it covers the entire graph. Make sure you freeze any external data that's needed for this operation and put it inside the assets.

### Definition of done
We have a suite of automated tests executed by `pytest`, which assure that the implementation of tram rail network transformation is successfully meeting all the requirements.

### Notes
The written tests should be unit tests. For inspiration, you can take a look at tests written in #7.
As for integration tests, these will be added in a separate PR.

PR #25, opened 2025-04-08
Codefident/Tram tracks transformer tests

Codefident:
### Related issues
- #11 

### Short description
Added tests for tram tracks graph transformer, covering correct graph interpolation, tracks crossings and handling incorrect `max_interpolation_distance`.


server/tests/tram_track_graph_transformer/test_tram_track_graph_transformer.py
```diff
@@ -0,0 +1,218 @@
...
+        with zip_file.open("osm_tram_stops_and_tracks.pickle") as file:
+            osm_tram_stops_and_tracks: overpy.Result = pickle.load(file)
+        return osm_tram_stops_and_tracks
+
+    def _get_tram_stops_data(self, zip_file: ZipFile) -> overpy.Result:
+        with zip_file.open("osm_tram_stops.pickle") as file:
+            osm_tram_stops: overpy.Result = pickle.load(file)
+        return osm_tram_stops
+
+    def test_if_tram_track_crossings_in_graph(
```
RCRalph, line 54: The convention for naming test methods is: `test_{method}_{short description - optional}`. You can take a look around the code base for inspiration

server/tests/tram_track_graph_transformer/test_tram_track_graph_transformer.py
```diff
@@ -0,0 +1,218 @@
...
+import overpy
+import pytest
+from geopy.distance import geodesic
+from networkx import DiGraph
+from src.model import CityConfiguration
+from src.tram_track_graph_transformer import TramTrackGraphTransformer
+from src.tram_track_graph_transformer.node import Node
+from tests.constants import FROZEN_DATA_DIRECTORY
+
+FILES_WITH_DATA = ["2025-04-22T03-35-34.zip"]
```
RCRalph, line 15: I think I forgot to tell you that the frozen data will be used only for integration tests, for tests of this class you should save an instance of correct data as a class and then manipulate it in the tests itself in order to get the results you wanted.

The assets you want to use should be located in `tests/assets` directory and be zipped with level 9 compression. You can then add that data as a fixture to `conftest.py`.

Please take a look at `TestTramStopGraphReachability` for inspiration.

server/dev-requirements.txt
```diff
@@ -1,3 +1,4 @@
 pytest>=8.3.4
 pytest-cov>=6.0.0
 httpx>=0.28.1
+geopy>=2.4.1
```
RCRalph, line 4: We're already using `pyproj` library for geographic calculations, so I think it's not necessary to introduce this dependency.

server/config/cities/krakow.json
RCRalph, line 1: I'm not sure about introducing these properties to the city configuration as these properties are only used in tests. Instead, I suggest introducing `tram_track_intersection_node_ids` and `tram_stop_node_ids` fixtures, which would read data from file(s) stored in `tests/assets` directory. These fixtures would contain the expected track intersection node IDs and tram stop node IDs, which should exist in the checked graph. Notice that in these tests we'll only be checking the single tram track graph configuration and it's OK, because we'll have frozen data and integration tests, which will cover the other cases.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -277,7 +277,35 @@ def densify_graph_by_max_distance(
                     path_nodes[-1],
                 )
 
-        if errors:
-            raise ExceptionGroup("Track direction errors during densification", errors)
+        # if errors:
```
RCRalph, line 280: I suspect you commented-out this code due to errors in the OSM data. Let's bring back this code even though it might fail the tests for now.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -277,7 +277,35 @@ def densify_graph_by_max_distance(
...
+        """
+        if id in self._nodes_by_id:
+            return self._nodes_by_id[id]
+        return None
+
+    def get_permament_nodes(self) -> set[Node]:
+        """
+        Get set of permament nodes from the graph
+        """
+        return self._permanent_nodes
```
RCRalph, line 305: I'm not sure when, but I think the `permanent_nodes` property has been removed from the `TramTrackGraphTransformer` class. You can use the `@property` decorator in order to achieve a better interface:
```suggestion
    @property
    def permament_nodes(self):
        return self._permanent_nodes.copy()
```
I'm not a huge fan of return type hints where it's not necessary, as the IDE should figure it out for you anyways and this way we'll avoid mistakes with changing the return type but not the type hint.

I'm also not a big fan of *obvious* docstrings, I believe that `permanent_nodes` contains enough information to know what this property contains.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -277,7 +277,35 @@ def densify_graph_by_max_distance(
...
+        ----------
+        id : int
+
+        Returns
+        ----------
+        Node or None
+        """
+        if id in self._nodes_by_id:
+            return self._nodes_by_id[id]
+        return None
```
RCRalph, line 299: As we have overloaded the `__eq__` and `__hash__` magic methods in `Node` class, we can now check if a given node is in the graph by just providing its ID, which means we shouldn't need this method anymore.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -277,7 +277,35 @@ def densify_graph_by_max_distance(
...
+        """
+        Get set of permament nodes from the graph
+        """
+        return self._permanent_nodes
+
+    def get_basic_graph(self) -> nx.DiGraph:
+        """
+        Get basic tram tracks graph
+        """
+        return self._tram_track_graph
```
RCRalph, line 311: Not sure if this method is used anywhere, could you check and remove if it's not used?

server/tests/_utils/util_overpass_client.py
```diff
@@ -0,0 +1,47 @@
+import overpy
+
+
+class UtilOverpassClient:
```
RCRalph, line 4: Not sure about that idea, but I'm thinking of creating a class `OverpassTestClient`, which would inherit from `OverpassClient` and add queries and class methods used only for the frozen data test cases. What do you think? This way we could eliminate the import of both `OverpassClient` and `UtilOverpassClient` inside the `add_frozen_data_test_case.py` file.

RCRalph (changes requested):

server/tests/conftest.py
```diff
@@ -41,3 +41,39 @@ def tram_trips_by_id() -> dict[str, list[int]]:
...
+        return json.load(file)
+
+
+@pytest.fixture
+def osm_tram_track_crossings() -> overpy.Result:
+    with zipfile.ZipFile(
+        "tests/assets/tram_track_crossings_overpass_query_result.zip"
+    ) as zip_file:
+        with zip_file.open("osm_tram_track_crossings.pickle") as file:
+            osm_tram_track_crossings: overpy.Result = pickle.load(file)
```
RCRalph, line 58: I think there's no need for type hints here, the one in the function signature should be enough:
```suggestion
            osm_tram_track_crossings = pickle.load(file)
```
(Applies to the entire file :3)

server/tests/tram_track_graph_transformer/test_tram_track_graph_transformer.py
```diff
@@ -0,0 +1,190 @@
...
+                not in krakow_ignored_crossings_and_stops["ignored_tram_stops_ids"]
+            ):
+
+                # assert
+                assert graph.has_node(node_id) is True
+
+    @pytest.mark.parametrize(
+        "max_densification_distance", CORRECT_MAX_DENSIFICATION_DISTANCES
+    )
+    def test_tram_track_transformer_densify_max_distance(
```
RCRalph, line 84: I think in this test it will be sufficient to just check all edges in this graph to see if their distance is not greater than `max_densification_distance`, it'll probably be a little simpler. that way.

server/tests/tram_track_graph_transformer/test_tram_track_graph_transformer.py
```diff
@@ -0,0 +1,190 @@
+from collections import deque
+from math import sqrt
+
+import overpy
+import pytest
+from networkx import DiGraph
+from pyproj import Geod
+from src.tram_track_graph_transformer import TramTrackGraphTransformer
+from src.tram_track_graph_transformer.node import Node
```
RCRalph, line 9: I think you can combine these two imports:
```suggestion
from src.tram_track_graph_transformer import TramTrackGraphTransformer, Node
```

server/tests/tram_track_graph_transformer/test_tram_track_graph_transformer.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+@pytest.fixture()
+def graph_25(osm_tram_stops_and_tracks, krakow_city_configuration):
+    transformer = TramTrackGraphTransformer(
+        osm_tram_stops_and_tracks, krakow_city_configuration
+    )
+    return (transformer, transformer.densify_graph_by_max_distance(25.0))
```
RCRalph, line 17: I'm not sure if we gain much from extracting this into a fixture, from my experience we shouldn't worry that much about repeated code as it's just two lines long. I suggest moving this to each of the tests, then it will become clear what parts are **Arrange**, **Act** and **Assert**,

server/tests/tram_track_graph_transformer/test_tram_track_graph_transformer.py
```diff
@@ -0,0 +1,190 @@
...
+                    prev_node = next_node
+                    next_node = list(densified_graph.successors(prev_node))[0]
+
+                n = len(distances)
+                if not n:
+                    continue
+
+                mean = sum(distances) / n
+                squares_sum = sum(map(lambda x: (x - mean) ** 2, distances))
+                sigma = sqrt(squares_sum / n)
```
RCRalph, line 164: I suggest extracting this part into a separate method, which would return `sigma` (whatever that is :3)

server/tests/tram_track_graph_transformer/test_tram_track_graph_transformer.py
```diff
@@ -0,0 +1,190 @@
...
+                visited.add(prev_node)
+
+                while next_node not in perm_nodes and next_node not in visited:
+                    _, _, distance = self._geod.inv(
+                        prev_node.lon, prev_node.lat, next_node.lon, next_node.lat
+                    )
+                    distances.append(distance)
+                    visited.add(prev_node)
+                    prev_node = next_node
+                    next_node = list(densified_graph.successors(prev_node))[0]
```
RCRalph, line 156: I think it would be nice if this was extracted into a separate method as well

server/tests/tram_track_graph_transformer/test_tram_track_graph_transformer.py
```diff
@@ -0,0 +1,190 @@
...
+        graph_25: tuple[TramTrackGraphTransformer, DiGraph],
+    ):
+        # arrange
+        _, graph = graph_25
+
+        for node_id in osm_tram_track_crossings.get_node_ids():
+            if (
+                node_id
+                not in krakow_ignored_crossings_and_stops["ignored_crossings_ids"]
+            ):
```
RCRalph, line 38: Maybe instead of excluding the nodes during runtime, we can prepare the data so it doesn't contain the nodes we want to skip. Take a look at: https://gis.stackexchange.com/questions/300896/exclude-elements-from-overpass-query, you can exclude nodes by their IDs in the query itself. If you need any help, ask ChatGPT, it's surprisingly good at Overpass queries :3

server/tests/tram_track_graph_transformer/test_tram_track_graph_transformer.py
```diff
@@ -0,0 +1,190 @@
...
+
+                # assert
+                assert graph.has_node(node_id) is True
```
RCRalph, line 41: I believe there's no need to be that specific:
```suggestion
                assert graph.has_node(node_id)
```
(Applies to the entire file)

RCRalph (commented):
Make sure to rebase the branch as there seem to be quite a few changes in conflict with this one

tests/tram_track_graph_transformer/test_tram_track_graph_transformer.py
```diff
@@ -0,0 +1,184 @@
...
+                    prev_node = next_node
+                    next_node = list(densified_graph.successors(prev_node))[0]
+
+                n = len(distances)
+                if not n:
+                    continue
+
+                mean = sum(distances) / n
+                squares_sum = sum(map(lambda x: (x - mean) ** 2, distances))
+                sigma = sqrt(squares_sum / n)
```
RCRalph, line 158: I suggest extracting this into a separate method

tests/tram_track_graph_transformer/test_tram_track_graph_transformer.py
```diff
@@ -0,0 +1,184 @@
...
+            succ_nodes = densified_graph.successors(perm_node)
+
+            for succ_node in succ_nodes:
+                if succ_node in perm_nodes:
+                    continue
+                distances = []
+                prev_node = perm_node
+                next_node = succ_node
+                visited = set()
+                visited.add(prev_node)
```
RCRalph, line 141: You can initialize a set like this:
```suggestion
                visited = {prev_node}
```

tests/tram_track_graph_transformer/test_tram_track_graph_transformer.py
RCRalph, line 1: Please add type hints to all methods in this file (even for the fixtures) and we will be done :3

RCRalph (approved):
Good job :3

Issue #19, opened 2025-04-05
Implement distance, azimuth and stop name assignment for graph's nodes and edges


RCRalph:
### Description
In order to allow the possibility of implementation of data structures presented in #16, the graph returned from the densification process should be supplemented with the following data:
- edge distance
- edge azimuth
- stop name for tram stops

### Definition of done
There is a function/method/class, which when provided with a graph with nodes of `Node` instances, supplies it with data mentioned above. Some test cases should be added as well, only after #11 gets merged.

### Notes
Some changes to the existing `TramTrackGraphTransformer` class should be made. For the attribute assignment for edges, I propose adding a method to the above mentioned class, but I'm open to implementing this functionality differently if it's somehow better.

Redor144:
Stop name for tram stop could add while constructing `self._node_by_id` by creating method sth like this:
```python
...
self._nodes_by_id = self._build_nodes_by_id_dict(
            tram_stops_and_tracks.get_nodes()
        )
...
def _build_nodes_by_id_dict(self, nodes: list[overpy.Node]) -> dict[int, Node]:
        nodes_by_id = {}
        for node in nodes:
            node_type = NodeType.get_by_value_safe(node.tags.get("railway"))
            nodes_by_id[node.id] = Node(
                id=node.id,
                lat=float(node.lat),
                lon=float(node.lon),
                type=node_type,
                name=node.tags.get("name") if node_type == NodeType.TRAM_STOP else None,
            )
        return nodes_by_id
```
The `Node` class will contain optional attribute `name: Optional[str] = None`.
I think the best place for edges attribute computing is right before we add them to the graph, what you think?

RCRalph:
> Stop name for tram stop could add while constructing `self._node_by_id` by creating method sth like this:
> 
> ...
> self._nodes_by_id = self._build_nodes_by_id_dict(
>             tram_stops_and_tracks.get_nodes()
>         )
> ...
> def _build_nodes_by_id_dict(self, nodes: list[overpy.Node]) -> dict[int, Node]:
>         nodes_by_id = {}
>         for node in nodes:
>             node_type = NodeType.get_by_value_safe(node.tags.get("railway"))
>             nodes_by_id[node.id] = Node(
>                 id=node.id,
>                 lat=float(node.lat),
>                 lon=float(node.lon),
>                 type=node_type,
>                 name=node.tags.get("name") if node_type == NodeType.TRAM_STOP else None,
>             )
[...]

PR #30, opened 2025-04-17
added atributes to edges and nodes

Redor144:
### Related issues
- #19

### Short description
Add `name` attribute to nodes if available; otherwise set to `None`. Azimuth and length of edges in the densified graph are now computed during their creation. No tests yet — pending initial feedback.

### DIFF
--- a/server/src/tram_track_graph_transformer/node.py
+++ b/server/src/tram_track_graph_transformer/node.py
@@ -1,3 +1,5 @@
+from typing import Optional
+
 from pydantic import BaseModel
 from src.tram_track_graph_transformer.node_type import NodeType
 
@@ -10,12 +12,14 @@ class Node(BaseModel, frozen=True):
         lat (float): Geographic latitude of the node.
         lon (float): Geographic longitude of the node.
         type (NodeType): Type of the node (e.g., tram stop, switch).
+        name (Optional[str]): The name of the tram stop, set only when type == NodeType.TRAM_STOP.
     """
 
     id: int
     lat: float
     lon: float
     type: NodeType
+    name: Optional[str] = None
 
     @property
     def coordinates(self):
--- a/server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
+++ b/server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
@@ -2,7 +2,7 @@ import math
 
 import networkx as nx
 import overpy
-from pyproj import Transformer
+from pyproj import Geod, Transformer
 from shapely.geometry import LineString
 from src.model import CityConfiguration
 from src.tram_track_graph_transformer.exceptions import (
@@ -43,6 +43,7 @@ class TramTrackGraphTransformer:
                 lat=float(node.lat),
                 lon=float(node.lon),
                 type=self._get_node_type(node),
+                name=node.tags.get("name"),
             )
             for node in tram_stops_and_tracks.get_nodes()
         }
@@ -200,6 +201,7 @@ class TramTrackGraphTransformer:
         last_node: Node,
     ):
         previous_graph_node = first_node
+        geod = Geod(ellps="WGS84")
 
         for lat, lon in interpolated_node_coordinates[1:-1]:
             if (lat, lon) in nodes_by_coordinates:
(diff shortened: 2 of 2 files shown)


olobuszolo (approved):

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -213,14 +215,30 @@ def _add_interpolated_nodes_path(
...
-            densified_graph.add_edge(previous_graph_node, new_node)
+            azimuth, _, length = geod.inv(
+                previous_graph_node.lon,
+                previous_graph_node.lat,
+                new_node.lon,
+                new_node.lat,
+            )
+            densified_graph.add_edge(
+                previous_graph_node, new_node, azimuth=azimuth, length=length
+            )
```
RCRalph, line 226: Actually, this code fragment is repeated 2 times, maybe it would look better if it was extracted to a separate method, what do you think?

Redor144: Yeah, that would improve legibility.

server/src/tram_track_graph_transformer/node.py
RCRalph, line 1: I think we previously missed setting `frozen=True` in class inheritance, which I think doesn't have any effect. The proper way of doing this in Pydantic is to add a `model_config` class property, like so:
```py
model_config = ConfigDict(frozen=True)
```

Probably the best place to put it is between the docstring and property definitions.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -191,6 +192,16 @@ def _get_new_node_id(self):
         self._max_node_id += 1
         return self._max_node_id
 
+    def _add_edge(self, graph: nx.DiGraph, first_node: Node, second_node: Node):
+        geod = Geod(ellps="WGS84")
```
RCRalph, line 196: Instead of creating `geod` on each method call, I suggest moving this to a protected class property.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -191,6 +192,16 @@ def _get_new_node_id(self):
         self._max_node_id += 1
         return self._max_node_id
 
+    def _add_edge(self, graph: nx.DiGraph, first_node: Node, second_node: Node):
```
RCRalph, line 195: This method doesn't use `self` anywhere, so it could be a class method.

I believe the method name could be more descriptive, as inside of it we're doing something more than just adding an edge, something like `_add_geodetic_edge` would be nice.

Also as we're dealing with a directed graph, I think a better name for these arguments would be something like `source_node` and `dest_node`, which would make it clear what will be the direction of the added edge. 

```suggestion
    def _add_geodetic_edge(self, graph: nx.DiGraph, source_node: Node, dest_node: Node):
```

server/src/tram_track_graph_transformer/node.py
Redor144: I don't think we even need that. In first implementation of class `Node` `frozen=True` was used for comparisons, but now `__eq__` and `__hash__` are defined manually.

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
Redor144: That's right. We can also move those two from `_interpolate_path_nodes` to protected class property. What you think?
```python
transformer = Transformer.from_crs("EPSG:4326", "EPSG:2180", always_xy=True)
inverse = Transformer.from_crs("EPSG:2180", "EPSG:4326", always_xy=True)
```

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
Redor144: `_interpolate_path_nodes` also doesn't use self. Maybe let it be class method too?

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
RCRalph: Let's do it

server/src/tram_track_graph_transformer/node.py
RCRalph: In this case let's make sure the `id` field is frozen, you can do this using Pydantic's `Field`

server/src/tram_track_graph_transformer/tram_track_graph_transformer.py
```diff
@@ -43,13 +47,54 @@ def __init__(
...
                 type=self._get_node_type(node),
+                name=node.tags.get("name"),
             )
             for node in tram_stops_and_tracks.get_nodes()
         }
         self._tram_track_graph = self._build_tram_track_graph_from_osm_ways()
         self._permanent_nodes = self._find_permanent_nodes()
         self._max_node_id = max(node.id for node in self._permanent_nodes)
 
+    @classmethod
```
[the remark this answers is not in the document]
Redor144: Yeah, thought that convection to have class methods right after `__init__` will be useful, but here they can stay in previous places.

RCRalph (approved):

Issue #33, opened 2025-04-30
Implement daily data updates


RCRalph:
### Description
In order to enable fast loading of data from the server, the process of building response data should happen at most once per day. The result of the data processing pipeline should then be saved and used until the next pipeline succeeds. In case the pipeline fails, the error message should be logged, but the server should continue working. The pipelines should be run for all available city configurations. Additionally, the response should be expanded with a field, which informs the user when was the last successful data update.

### Definition of done
When sending a request to `/cities/{city_id}` endpoint, calculating the response doesn't require running the full data processing pipeline. The data sent by the server is refreshed as most once per day. The data processing errors can be observed in the logs. The last successful data update should be available in the response body.

### Notes
Not sure if tests for this functionality are possible, please add them if they are.

PR #36, opened 2025-05-08
Add caching for city response data with fallback on failure

Redor144:
### Related issues
- #33

### Short description
Added caching mechanism for the `/cities/{city_id}` endpoint to improve response time.
Introduced the `CityDataCache` class for managing cache storage.

### DIFF
--- a/src/city_data_cache.py
+++ b/src/city_data_cache.py
@@ -2,6 +2,16 @@ import json
 from datetime import datetime, timedelta
 from pathlib import Path
 
+from pydantic import BaseModel
+
+from src.city_data_builder.model import ResponseGraphNode, ResponseTramTrip
+
+
+class ResponseCityData(BaseModel):
+    tram_track_graph: list[ResponseGraphNode]
+    tram_trips: list[ResponseTramTrip]
+    last_updated: str = datetime.now().isoformat()
+
 
 class CityDataCache:
     def __init__(
@@ -22,25 +32,29 @@ class CityDataCache:
             return False
         return datetime.now() - datetime.fromtimestamp(path.stat().st_mtime) < self.ttl
 
-    def load_cached_data(self, city_id: str) -> dict:
+    def load_cached_data(self, city_id: str) -> ResponseCityData:
         with open(self._get_path_to_cache(city_id), "r", encoding="utf-8") as f:
             data = json.load(f)
-        data["last_updated"] = self._get_last_modified_timestamp(city_id)
-        return data
+        return ResponseCityData(**data)
 
-    def save_to_cache(self, city_id: str, data: dict):
+    def store_and_return(
+        self,
+        city_id: str,
+        tram_track_graph: list[ResponseGraphNode],
+        tram_trips: list[ResponseTramTrip],
+    ) -> ResponseCityData:
         def default(obj):
             if hasattr(obj, "model_dump"):
                 return obj.model_dump(mode="json")
             raise TypeError(f"{type(obj).__name__} is not JSON serializable")
 
+        data = ResponseCityData(
+            tram_track_graph=tram_track_graph,
+            tram_trips=tram_trips,
+            last_updated=datetime.now().isoformat(),
+        )
+
         with open(self._get_path_to_cache(city_id), "w", encoding="utf-8") as f:
             json.dump(data, f, default=default)
 
-    def _get_last_modified_timestamp(self, city_id: str) -> str | None:
-        path = self._get_path_to_cache(city_id)
-        return (
-            datetime.fromtimestamp(path.stat().st_mtime).isoformat()
-            if path.exists()
-            else None
-        )
+        return data
--- a/src/city_data_cache.py
+++ b/src/city_data_cache.py
@@ -3,11 +3,15 @@ from pathlib import Path
 
 from pydantic import BaseModel, Field
 
-from src.city_data_builder.model import ResponseGraphNode, ResponseTramTrip
+from src.city_data_builder.model import (
+    ResponseGraphNode,
+    ResponseGraphTramStop,
+    ResponseTramTrip,
+)
 
 
 class ResponseCityData(BaseModel):
-    tram_track_graph: list[ResponseGraphNode]
+    tram_track_graph: list[ResponseGraphNode | ResponseGraphTramStop]
     tram_trips: list[ResponseTramTrip]
     last_updated: datetime = Field(default_factory=lambda: datetime.now())
 
--- a/src/city_data_cache.py
+++ b/src/city_data_cache.py
@@ -1,8 +1,7 @@
-import json
 from datetime import datetime, timedelta
 from pathlib import Path
 
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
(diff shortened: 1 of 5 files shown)


src/server.py
```diff
@@ -39,13 +42,35 @@ def get_city_data(city_id: str):
...
-    city_data_builder = CityDataBuilder(city_configuration)
-
-    return {
-        "tram_track_graph": city_data_builder.tram_track_graph_data,
-        "tram_trips": city_data_builder.tram_trips_data,
-    }
+    if cache.is_valid(city_id):
+        logger.info(f"Loading data from cache for {city_id}")
+        city_data = cache.load(city_id)
+        city_data["last_updated"] = cache.last_updated(city_id)
```
RCRalph, line 48: I suggest adding this field to the result of `CityDataCache.load` method instead of adding it manually every time it's loaded or saved.

src/server.py
```diff
@@ -39,13 +42,35 @@ def get_city_data(city_id: str):
...
+        response_data = {
+            "tram_track_graph": city_data_builder.tram_track_graph_data,
+            "tram_trips": city_data_builder.tram_trips_data,
+            "last_updated": datetime.now().isoformat(),
+        }
+        cache.save(city_id, response_data)
+        return response_data
+
+    except Exception as exc:
+        logger.exception(f"Failed to build city data for {city_id}: {exc}")
```
RCRalph, line 63: When logging an exception, you can pass the `exc` as a keyword argument in order to provide it with more context:
```suggestion
        logger.exception(f"Failed to build city data for {city_id}", exc_info=exc)
```

src/server.py
```diff
@@ -39,13 +42,35 @@ def get_city_data(city_id: str):
...
+
+        if (CONFIG_DIRECTORY_PATH / f"{city_id}.json").exists():
+            try:
+                cached_data = cache.load(city_id)
+                cached_data["last_updated"] = cache.last_updated(city_id)
+                return cached_data
+            except Exception as inner:
+                logger.exception(f"Cache loading failed for {city_id}: {inner}")
+
+        raise HTTPException(500, "Data unavailable for {city_id}")
```
RCRalph, line 73: Not sure if 500 is the correct status code here, it does sound like an issue with the base data and not the server.

src/city_data_cache.py
```diff
@@ -0,0 +1,44 @@
+import json
+from datetime import datetime, timedelta
+from pathlib import Path
+
+
+class CityDataCache:
```
RCRalph, line 6: I believe method names in this class could be a bit more descriptive.

src/city_data_cache.py
```diff
@@ -0,0 +1,44 @@
...
+            return json.load(f)
+
+    def save(self, city_id: str, data: dict):
+        def default(obj):
+            if hasattr(obj, "model_dump"):
+                return obj.model_dump(mode="json")
+            raise TypeError(f"{type(obj).__name__} is not JSON serializable")
+
+        with open(self._get_path(city_id), "w", encoding="utf-8") as f:
+            json.dump(data, f, indent=4, default=default)
```
RCRalph, line 36: No need for pretty-formatting here, we're not reading these files manually anyways

RCRalph (changes requested):

src/server.py
```diff
@@ -39,13 +42,30 @@ def get_city_data(city_id: str):
...
-        "tram_trips": city_data_builder.tram_trips_data,
-    }
+    try:
+        city_configuration = CityConfiguration.from_path(file_path)
+        city_data_builder = CityDataBuilder(city_configuration)
+        response_data = {
+            "tram_track_graph": city_data_builder.tram_track_graph_data,
+            "tram_trips": city_data_builder.tram_trips_data,
+            "last_updated": datetime.now().isoformat(),
+        }
```
RCRalph, line 56: Let's maybe introduce a Pydantic model which would hold the response data, this would automatically enhance the endpoint documentation as well as simplify a few things.

RCRalph: While you're at it, you can add the return type hint to the `cities` function which should enhance the documentation.

tests/test_server.py
```diff
@@ -171,3 +172,35 @@ def test_get_city_data_unknown_city_id(self):
         # Assert
         assert response.status_code == 404
         assert response.json()["detail"] == "City not found"
+
+    @patch("src.server.cache")
+    @patch("src.server.CityDataBuilder.__init__", return_value=None)
+    @patch("src.server.CityDataBuilder.tram_track_graph_data", new_callable=MagicMock)
+    @patch("src.server.CityDataBuilder.tram_trips_data", new_callable=MagicMock)
+    def test_cache_used_after_initial_build(
```
RCRalph, line 180: I believe unit tests should cover the `CityDataCache` class itself rather than the endpoint using it

tests/test_city_cache_data.py
```diff
@@ -0,0 +1,61 @@
...
+        tram_track_graph = [
+            ResponseGraphNode(
+                id=1,
+                lat=50.0,
+                lon=19.0,
+                neighbors=[ResponseGraphEdge(id=2, length=100.0, azimuth=90.0)],
+            )
+        ]
+        tram_trips = [ResponseTramTrip(route="1", trip_head_sign="Centrum", stops=[])]
+        return tram_track_graph, tram_trips
```
RCRalph, line 33: This is a fixture

tests/test_city_cache_data.py
```diff
@@ -0,0 +1,61 @@
...
+    ResponseTramTrip,
+)
+from src.city_data_cache import CityDataCache, ResponseCityData
+
+
+class TestCityDataCache:
+    @pytest.fixture
+    def temp_cache(self):
+        cache_dir = Path(tempfile.mkdtemp())
+        return CityDataCache(cache_dir=cache_dir, ttl_hours=1)
```
RCRalph, line 20: I think this code should be included in each test case, there's very little gain in terms of readability for moving this into a fixture.

tests/test_city_cache_data.py
```diff
@@ -0,0 +1,61 @@
...
+
+from src.city_data_builder.model import (
+    ResponseGraphEdge,
+    ResponseGraphNode,
+    ResponseTramTrip,
+)
+from src.city_data_cache import CityDataCache, ResponseCityData
+
+
+class TestCityDataCache:
```
RCRalph, line 16: Please add Arrange - Act - Assert comments so it's clear which part of the test does what thing

tests/test_city_cache_data.py
```diff
@@ -0,0 +1,61 @@
...
+                id=1,
+                lat=50.0,
+                lon=19.0,
+                neighbors=[ResponseGraphEdge(id=2, length=100.0, azimuth=90.0)],
+            )
+        ]
+        tram_trips = [ResponseTramTrip(route="1", trip_head_sign="Centrum", stops=[])]
+        return tram_track_graph, tram_trips
+
+    def test_cache_miss_if_no_file(self, temp_cache: CityDataCache):
```
RCRalph, line 35: Test method names should follow convention: `test_{tested method}_{short description unless default behaviour}`

RCRalph (approved):

olobuszolo (approved):

Issue #37, opened 2025-05-16
Recent changes in tram schedules cause a failure when sending a request to `/cities/krakow`


RCRalph:
### Description
When building response data for Kraków, the following error appears. It seems to be caused by the existence of trip between *Bardosa* and *Kopiec Wandy* stops, where *Bardosa* stop is mapped to *Bardosa 02*. This causes the route to be too long, because the shortest way to get between **Bardosa 02** and **Kopiec Wandy 02** is through *Plac Centralny im. Ronalda Reagana* intersection. 

```
Failed to build city data for krakow
Traceback (most recent call last):
  File "/home/rcralph/Documents/AGH/TNSEngineerEdition/Server/src/server.py", line 50, in get_city_data
    city_data_builder = CityDataBuilder(city_configuration)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rcralph/Documents/AGH/TNSEngineerEdition/Server/src/city_data_builder/city_data_builder.py", line 29, in __init__
    self._tram_track_graph = self._get_tram_track_graph()
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rcralph/Documents/AGH/TNSEngineerEdition/Server/src/city_data_builder/city_data_builder.py", line 66, in _get_tram_track_graph
    tram_track_graph_inspector.check_path_viability(
  File "/home/rcralph/Documents/AGH/TNSEngineerEdition/Server/src/tram_track_graph_transformer/tram_track_graph_inspector.py", line 65, in check_path_viability
    raise PathTooLongError(
src.tram_track_graph_transformer.exceptions.PathTooLongError: Path too long: 773635753 -> 2426893478 length: 4888.2 > allowed: 1304.1
```

### Expected outcome
Tram stop mapping completes successfully

### Actual outcome
Invalid tram stop mapping is created and `PathTooLongError` is raised.

PR #38, opened 2025-05-16
Added custom stop pair mapping and advanced custom stop mapping

RCRalph:
### Related issues
- #37

### Short description
Added ability to set custom tram stop mapping for first, general and last tram stop mappings using a three-element list.
This is used for specifying start and end stops for trams arriving or departing tram depots with multiple exits.

Also added possibility to specify the mapping for custom tram stop pair.
This is used for edge-cases when no relation covers a given trip and mapped tram stops are in the wrong way as a result.


olobuszolo (approved):
LGTM :)

Issue #39, opened 2025-05-21
Add grouping of tram trips by day of the week


RCRalph:
### Description
In GTFS Schedule dataset, tram trips have a `service_id` parameter. This parameter is defined in `calendar.txt` file and specifies the days of the week when the tram trips are happening. Currently we don't group tram trips by this parameter, but instead merge them together into one day, which duplicates tram trips. In order to avoid this, given city data should contain tram trips from only one day of the week.

The chosen day of the week should be provided by the API. There should be a `weekday` query parameter added to the GET `/cities/{city_id}` endpoint, which would accept strings of days of the week (`monday`, `tuesday`, ...). This parameter should be optional. When it's missing, the data should contain trips from the current weekday. 

### Definition of done
City data contains tram trips from only one weekday. The weekday can be chosen using the `weekday` query parameter in the API. By default, the current weekday is chosen.

PR #40, opened 2025-05-29
Added possibility to choose tram schedule for specific day, by defalt the day is chosen current day

olobuszolo:
### Related issues
- #39 

### Short description
Added the possibility to select the schedule based on the weekday. By default, the day corresponding to the current date is selected. The city configuration caching mechanism was also updated- cached data is now differentiated by weekday to support varying schedules.


src/city_data_builder/weekday_enum.py
```diff
@@ -0,0 +1,11 @@
+from enum import StrEnum
+
+
+class WeekdayEnum(StrEnum):
```
RCRalph, line 4: Let's just rename this to `Weekday` (applies to the file as well). Also please add a classmethod for getting the current weekday.

src/server.py
```diff
@@ -37,32 +42,42 @@ def cities() -> dict[str, CityConfiguration]:
 
 
 @app.get("/cities/{city_id}")
-def get_city_data(city_id: str) -> ResponseCityData:
+def get_city_data(
+    city_id: str, weekday: str | None = datetime.datetime.now().strftime("%A").lower()
```
RCRalph, line 46: The current configuration will every time default to the weekday when the endpoint was first used. Instead, add a check if the weekday is specified.

src/server.py
```diff
@@ -37,32 +42,42 @@ def cities() -> dict[str, CityConfiguration]:
...
+) -> ResponseCityData:
     if not (file_path := CONFIG_DIRECTORY_PATH / f"{city_id}.json").is_file():
         raise HTTPException(404, "City not found")
+    city_daily_schedule = f"{city_id}_{weekday}"
```
RCRalph, line 50: It's probably a good idea to move this variable definition to the place where it's needed the first time (1 if statement below)

src/tram_stop_mapper/gtfs_package.py
RCRalph, line 1: Please also update the `test_gtfs_package` test with checks for the calendar DataFrame.

src/tram_stop_mapper/gtfs_package.py
```diff
@@ -48,6 +58,24 @@ def from_url(cls, url: str):
...
+    def _service_id_as_dict(self) -> dict[str, set[str]]:
+        days = [
+            "monday",
+            "tuesday",
+            "wednesday",
+            "thursday",
+            "friday",
+            "saturday",
+            "sunday",
+        ]
```
RCRalph, line 71: You can get the same list using `list(Weekday)`

src/tram_stop_mapper/gtfs_package.py
```diff
@@ -48,6 +58,24 @@ def from_url(cls, url: str):
     def _stop_times_as_dict(self) -> dict[str, dict[tuple[str, int], Any]]:
         return self.stop_times.set_index(["trip_id", "stop_sequence"]).to_dict()
 
+    @cached_property
+    def _service_id_as_dict(self) -> dict[str, set[str]]:
```
RCRalph, line 62: Type hint isn't equivalent to what's actually being returned

src/tram_stop_mapper/gtfs_package.py
```diff
@@ -67,7 +95,7 @@ def _time_string_to_seconds(time_str: str) -> int:
     def trip_data_and_stops_by_trip_id(self):
         stop_ids = self._stop_times_as_dict["stop_id"]
         departure_times = self._stop_times_as_dict["departure_time"]
-
+        service_id_to_days = self._service_id_as_dict
```
RCRalph, line 98: I think there's no need to save it to a separate variable (it's a cached property)

src/tram_stop_mapper/gtfs_package.py
```diff
@@ -83,9 +111,13 @@ def trip_data_and_stops_by_trip_id(self):
...
         for trip_id, trip_row in self.trips.iterrows():
+            service_id = trip_row["service_id"]
             for name, value in trip_row.items():
-                trip_data_by_trip_id[trip_id][name] = value
-
+                if name != "service_id":
+                    trip_data_by_trip_id[trip_id][name] = value
+            trip_data_by_trip_id[trip_id]["service_days"] = list(
+                service_id_to_days.get(service_id)
+            )
```
RCRalph, line 120: Why is this a list instead of a set?

RCRalph, line 120: Maybe something like this?
```suggestion
                if name == "service_id":
                    trip_data_by_trip_id[trip_id]["service_days"] = list(
                        service_id_to_days.get(service_id)
                    )
                else:
                    trip_data_by_trip_id[trip_id][name] = value
```

src/city_data_builder/weekday_enum.py
RCRalph, line 1: Are we sure this is the correct place for this enum? I would actually put it inside the `tram_stop_mapper` package instead.

RCRalph (changes requested):

src/tram_stop_mapper/gtfs_package.py
olobuszolo: Good point, changed to set

src/city_data_builder/weekday_enum.py
olobuszolo: The `Weekday` class now includes a `get_current_day` class method, which returns either the provided day or the current simulation day. The term 'current day' refers to the day for which we want to run the simulation, and it may differ from the actual calendar day. What do you think about this approach?

src/tram_stop_mapper/weekday.py
```diff
@@ -0,0 +1,22 @@
...
+
+    @classmethod
+    def get_current_weekday(cls, weekday: str | None = None):
+        if weekday is None:
+            return cls(datetime.datetime.now().strftime("%A").lower())
+
+        try:
+            return cls(weekday.lower())
+        except ValueError:
+            raise ValueError(f"Invalid weekday: {weekday}. Must be one of {list(cls)}")
```
RCRalph, line 22: This method shouldn't accept an argument, it should just return the case when weekday is `None`. As for conversion from `str` to `Weekday`, you can create a separate method for it like `get_by_value`.

RCRalph (approved):

Issue #47, opened 2025-08-22
Add endpoints for fetching available cached data


Codefident:
### Description
Cached data represent the processed graph and all the information required to run a simulation.
We need to expose this cached data through two endpoints so that the client can easily query and retrieve it:

1. **List available data** - an endpoint that returns basic information about available data (it can be presented as dates, like `25.05.2025, 10. 07.2025, 21.08.2025`, which means that cached data for each of these days is available).
2. **Fetch data for a specific day** - an endpoint which returns all the data for a specific day.

### Definition of done
Both endpoints exists and return the data correctly.

### Notes
We might consider adding a pagination in case the list of cached data gets too long (although it probably won't 🤗)

Codefident:
#

> Could you be more specific with the API design? I'm especially after the exact routes and parameters.

I haven't really thought of any exact routes yet. When it comes to the parameters I think we might use the city name (e.g. `krakow`) and some ID of a data (this ID may be a coded date, e.g. `220825` or `22_05_2025`).

I will try to come up with the exact routes and parameters, but if you have any idea feel free to share it 😅

#

> Also should we think about limiting the maximum amount of cached data, to for example 10 per each city?
> This could help us avoid issues with cache invalidation and storage space.

That's a good idea. Should we create a new issue for this as it's more related to caching the data, not fetching it?

PR #49, opened 2025-09-11
Olobuszolo/endpoints available cached data

olobuszolo:
### Related issues
- #47 

### Short description
Changed endpoints

1. **`GET /cities`**  
   Returns:  
   - the latest `CityConfiguration`  
   - all `available_dates`  

   for all cities that have cached data.  

2. **`GET /cities/<city_id>`**  
   Returns data for a specific city.  

   **Query parameters:**  
   - `date` – the configuration date to retrieve (defaults to **today**)  
   - `weekday` – the GTFS weekday (defaults to the weekday of the given date)  
---
   **Notes:**  
The way how server caches/returns data is compatible with Rafal's description: 
1. Query parameters aren't provided
If the cache for today's date is missing, the server builds the data using newest available configuration using the today's weekday. If the build is successful, the cache is stored under today's date. The server then returns the cached data for today's date.

2. Date is provided
Server checks if the cache for provided date is available. If it is, the server returns the cached data, otherwise it raises a 404 error.

3. Weekday is provided
The server builds the data using newest available configuration using the provided weekday, downloads the GTFS dataset, and runs Overpass queries. If the build is successful, the data is not cached and is immediately returned.

4. Both arguments are provided
The server returns 400 error.
 ---
### Changed cache directory

The cache directory now has the following structure:

cache/
├──cities/
        ├── krakow/
          │   ├── 2025-01-18.json
          │   ├── 2025-06-13.json
          │   └── 2025-07-25.json
          └── warsaw/
              ├── 2025-03-05.json
              ├── 2025-05-31.json
              └── 2025-07-01.json


src/city_data_cache/city_data_cache.py
```diff
@@ -20,37 +19,63 @@ def __init__(
...
+        cache_file_path = self._get_path_to_cache(city_id, date)
         if not cache_file_path.is_file():
-            return False
+            return None
 
-        timedelta_since_last_update = datetime.now() - datetime.fromtimestamp(
-            cache_file_path.stat().st_mtime
+        return ResponseCityData.model_validate_json(
+            cache_file_path.read_text(encoding="utf-8")
         )
```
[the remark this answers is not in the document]
olobuszolo: Ok, done, now `date` has type `datetime.date` and we are converting it to string only when needed

src/city_data_cache/city_data_cache.py
```diff
@@ -20,37 +19,63 @@ def __init__(
...
+            result[city_id] = {
+                "city_configuration": city_config,
+                "available_dates": cached_dates,
+            }
 
-        return ResponseCityData.model_validate_json(cache_file_path.read_text())
+        return result
 
-    def store(
+    def build_and_store(
```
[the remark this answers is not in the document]
olobuszolo: Done

src/city_data_cache/city_data_cache.py
```diff
@@ -13,44 +11,63 @@ def __init__(
         cache_directory: Path = Path(
             os.environ.get("CITY_DATA_CACHE_DIRECTORY", "./cache/cities")
         ),
-        ttl_timedelta: timedelta = timedelta(hours=24),
+        max_file_count: int = 10,
```
RCRalph, line 14: I would like to set the default parameter of this value from an environment variable if possible

src/city_data_cache/city_data_cache.py
```diff
@@ -13,44 +11,63 @@ def __init__(
...
+            return None
 
-        timedelta_since_last_update = datetime.now() - datetime.fromtimestamp(
-            cache_file_path.stat().st_mtime
+        return ResponseCityData.model_validate_json(
+            cache_file_path.read_text(encoding="utf-8")
         )
 
-        return timedelta_since_last_update < self.ttl_timedelta
+    def get_cached_dates(self, city_id: str) -> list[datetime.date]:
```
RCRalph, line 39: The entire logic of this method could fit into a single list comprehension

src/city_data_cache/city_data_cache.py
```diff
@@ -13,44 +11,63 @@ def __init__(
...
 
-        return ResponseCityData.model_validate_json(cache_file_path.read_text())
+    def _remove_the_oldest_one(self, city_cache_dir: Path) -> None:
+        oldest_file = min(
+            cached_file
+            for cached_file in city_cache_dir.iterdir()
+            if cached_file.is_file() and cached_file.suffix == ".json"
+        )
+
+        oldest_file.unlink()
```
RCRalph, line 55: How about we remove the oldest files until there are `max_file_count-1` files left? It could be a bit more flexible this way.

src/server.py
```diff
@@ -17,36 +18,89 @@
...
+    try:
+        return datetime.datetime.strptime(date, "%Y-%m-%d").date()
+    except ValueError:
+        raise HTTPException(
+            status_code=400,
+            detail=f"Invalid date format: '{date}', expected YYYY-MM-DD",
+        )
+
+
+def validate_weekday(weekday: str | None = None) -> Weekday | None:
```
RCRalph, line 33: How about we handle the case when `weekday` is `None` here as well? We won't need to check for that case in `get_by_value` anymore.

src/server.py
```diff
@@ -17,36 +18,89 @@
...
-            logger.exception(
-                f"Failed to build city data for city {city_id} and weekday {weekday_enum}",
-                exc_info=exc,
-            )
-
-    if city_data := city_data_cache.get(city_id, weekday_enum):
+        raise HTTPException(404, f"City {city_id} not found")
+
+    try:
+        city_data_builder = CityDataBuilder(city_configuration, weekday)  # type: ignore[arg-type]
```
RCRalph, line 94: This line is the only line which could raise an exception, maybe it's a good idea to wrap only this line in try-except?

src/city_data_cache/model.py
```diff
@@ -1,15 +1,9 @@
...
-    ResponseGraphTramStop,
-    ResponseTramRoute,
-)
 
-
-class ResponseCityData(BaseModel):
-    tram_track_graph: list[ResponseGraphNode | ResponseGraphTramStop]
-    tram_routes: list[ResponseTramRoute]
-    last_updated: datetime = Field(default_factory=datetime.now)
+class CachedCityDates(TypedDict):
```
RCRalph, line 7: I suggest using `BaseModel` from Pydantic instead of `TypedDict`. We're using FastAPI which is already setup to handle `BaseModel` instances.

src/server.py
```diff
@@ -17,36 +18,89 @@
...
+
+
 @app.get("/cities")
-def cities() -> dict[str, CityConfiguration]:
+def cities() -> dict[str, CachedCityDates]:
     try:
-        return CityConfiguration.get_all()
+        result: dict[str, CachedCityDates] = {}
+
+        for city_id in city_data_cache.list_cities():
```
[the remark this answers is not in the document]
olobuszolo: No, currently when city hasn't got any cached data we return city configuration with empty list with available dates. 
image

src/server.py
```diff
@@ -17,36 +18,89 @@
...
+            city_config = CityConfiguration.get_by_city_id(city_id)
+            if city_config is None:
+                continue
```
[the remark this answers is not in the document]
olobuszolo: Added `any(city_directory.iterdir())` to `get_all` and this handle this case

src/server.py
```diff
@@ -17,38 +18,101 @@
...
+    city_id: str,
+    weekday: Weekday | None = Depends(validate_weekday),
+    date: datetime.date | None = Depends(validate_date),
+) -> ResponseCityData:
+    """
+    Returns processed data for a given city.
+
+    - Without parameters: returns data for the current day
+    (from cache if available, otherwise built and cached).
+    - With `weekday`: builds and returns data for the given day of the week.
```
RCRalph, line 73:
```suggestion
    - With `weekday`: builds and returns data for the given day of the week using currently available GTFS Schedule and OpenStreetMap data.
```

src/server.py
```diff
@@ -17,38 +18,101 @@
...
+
+    assert weekday is not None
+    assert date is not None
+
+    try:
+        city_data_builder = CityDataBuilder(city_configuration, weekday)
+    except Exception as exc:
+        logger.exception(
+            f"Failed to build city data for city {city_id} for weekday {weekday}",
+            exc_info=exc,
```
RCRalph, line 108: The recommended approach is to use lazy formatting for log messages (due to formatting being expensive so we want to do it only if we actually want to display this log message).
```suggestion
            "Failed to build city data for city %s for weekday %s",
            city_id,
            weekday
            exc_info=exc,
```

src/server.py
```diff
@@ -17,38 +18,101 @@
...
+    if date is not None and weekday is not None:
+        raise HTTPException(400, "Provide either date or weekday")
+
+    if date:
+        if cached_data := city_data_cache.get(city_id, date):
+            return cached_data
+
+        raise HTTPException(
+            404, f"City data for {city_id} not found in cache for {date}"
+        )
```
RCRalph, line 87: Just a thought - maybe it's a good idea to put these 3 different scenarios (no parameters, only weekday, only date) in separate methods? Maybe it will clean things up a bit, though I worry a little bit about code duplication. Could you try doing this to see if it makes sense?

src/city_data_cache/city_data_cache.py
```diff
@@ -1,56 +1,82 @@
...
+            reverse=True,
+        )
 
-    def get(self, city_id: str, weekday: Weekday) -> ResponseCityData | None:
-        cache_file_path = self._get_path_to_cache(city_id, weekday)
-        if not cache_file_path.is_file():
-            return None
+    def _remove_redundant_files(self, city_cache_dir: Path) -> None:
+        if not city_cache_dir.exists():
+            return
```
RCRalph, line 52: On second thought, I think it might be a little cleaner to check this condition outside of this method and create the directory if it doesn't exist.

src/city_data_cache/city_data_cache.py
```diff
@@ -1,56 +1,82 @@
...
+from city_data_builder import ResponseCityData
 
 
 class CityDataCache:
+    DEFAULT_CACHE_DIRECTORY: Path = Path(
+        os.environ.get("CITY_DATA_CACHE_DIRECTORY", "./cache/cities")
+    )
+    DEFAULT_MAX_FILE_COUNT: int = int(
+        os.environ.get("CITY_DATA_CACHE_MAX_FILE_COUNT", "10")
+    )
```
RCRalph, line 14: Not sure if the type hints are necessary here

src/server.py
olobuszolo: Yeah, I think your suggestion is a great idea, and I don’t see any problem with code duplication. Take a look and let me know what you think

src/server.py
```diff
@@ -17,38 +18,89 @@
...
-            )
-
-    if city_data := city_data_cache.get(city_id, weekday_enum):
-        return city_data
-
-    raise HTTPException(500, f"Data processing for {city_id} failed")
+        raise HTTPException(404, f"City {city_id} not found")
+
+    try:
+        city_data_builder = CityDataBuilder(city_configuration, weekday)  # type: ignore[arg-type]
```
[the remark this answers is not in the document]
olobuszolo: Now we dont't use this

RCRalph (commented):
It's looking really good now, final comments

olobuszolo:
It’s probably a good idea, but could we make it a separate, smaller task? I need to finish these tests and I’d rather not stretch this issue out too much over time. @RCRalph ???

RCRalph:
> It’s probably a good idea, but could we make it a separate, smaller task? I need to finish these tests and I’d rather not stretch this issue out too much over time. @RCRalph ???

Sure, I'll add this to our backlog and once we ship your changes we will assess if it's worth investing into.

pyproject.toml
```diff
@@ -56,3 +56,4 @@ force_alphabetical_sort_within_sections = true
 [tool.mypy]
 ignore_missing_imports = true
 strict = true
+mypy_path = "src"
```
RCRalph, line 59: What's the reason for this change?

tests/conftest.py
```diff
@@ -81,3 +82,15 @@ def krakow_response_city_data() -> ResponseCityData:
...
+@pytest.fixture(scope="class")
+def city_cache_dir() -> Generator[Path, Any, None]:
+    directory_path = Path(tempfile.mkdtemp())
+
+    with zipfile.ZipFile("tests/assets/cached_data.zip") as zip_file:
+        zip_file.extractall(directory_path)
+
+    yield directory_path / "cached_data"
+
+    shutil.rmtree(directory_path)
```
RCRalph, line 96: You can use `with tempfile.NamedDirectory` or something like that to automatically remove the directory when tests finish

tests/test_server.py
```diff
@@ -66,9 +70,12 @@ def test_cities(
                 for osm_node_id in custom_stop_mapping.values()
             )
 
+            assert isinstance(available_dates, list)
+            assert all(isinstance(d, str) for d in available_dates)
```
RCRalph, line 74: I would also like to see an assertion for if all items can be parsed into dates and that they are in the correct order

tests/test_server.py
```diff
@@ -172,15 +179,16 @@ def _assert_city_data_content(self, city_data: Any) -> None:
...
     @patch("overpass_client.OverpassClient.get_tram_stops_and_tracks")
     @patch("overpass_client.OverpassClient.get_relations_and_stops")
-    @patch("city_data_cache.CityDataCache.is_fresh", return_value=False)
+    @patch("server.CityDataCache.get", return_value=None)
+    @patch("server.CityDataCache.store", return_value=None)
     def test_get_city_data(
         self,
-        cache_is_fresh_mock: MagicMock,
+        store_mock: MagicMock,
+        cache_get_mock: MagicMock,
```
RCRalph, line 191: If you introduce these mocks, you should confirm that you're calling them and with what arguments

tests/test_server.py
```diff
@@ -218,22 +225,25 @@ def test_get_city_data(
...
 
+    @pytest.mark.parametrize("params", [{"weekday": "monday"}, {"date": "2025-01-01"}])
     @patch("city_data_builder.city_configuration.CityConfiguration.get_by_city_id")
     @patch("tram_stop_mapper.gtfs_package.GTFSPackage.from_url")
     @patch("overpass_client.OverpassClient.get_tram_stops_and_tracks")
     @patch("overpass_client.OverpassClient.get_relations_and_stops")
-    @patch("city_data_cache.CityDataCache.is_fresh", return_value=False)
-    def test_get_city_data_with_weekday(
+    @patch("server.CityDataCache.get")
+    def test_get_city_data_with_weekday_od_date(
```
RCRalph, line 234: I think these two tests could be separate

tests/conftest.py
olobuszolo: Now we're using 'tempfile.TemporaryDirectory()'

tests/test_server.py
```diff
@@ -285,65 +307,47 @@ def test_get_city_data_invalid_weekday(self) -> None:
...
-        krakow_response_city_data: ResponseCityData,
-        krakow_city_configuration: CityConfiguration,
-        caplog: pytest.LogCaptureFixture,
-    ) -> None:
+    def test_get_city_data_invalid_date(self) -> None:
         # Arrange
-        expected_log_message = (
-            "Failed to build city data for city krakow and weekday wednesday"
+        expected_response_detail = (
+            "Invalid date format: '2137-2137-2137', expected YYYY-MM-DD"
```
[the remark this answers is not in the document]
olobuszolo: So you probably wouldn’t like to hear my second idea for these params... Maybe the new boring ones will be `2025-13-13`

src/city_data_cache/city_data_cache.py
```diff
@@ -1,56 +1,81 @@
...
+                if city_by_date.is_file() and city_by_date.suffix == ".json"
+            ],
+            reverse=True,
+        )
 
-    def get(self, city_id: str, weekday: Weekday) -> ResponseCityData | None:
-        cache_file_path = self._get_path_to_cache(city_id, weekday)
-        if not cache_file_path.is_file():
-            return None
+    def _remove_redundant_files(self, city_cache_dir: Path) -> None:
```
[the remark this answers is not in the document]
olobuszolo: Created `test_remove_redundant_files`

tests/test_city_cache_data.py
```diff
@@ -54,52 +53,60 @@ def cache_directory(self) -> Generator[Path, Any, None]:
...
-        cache = CityDataCache(cache_directory, timedelta(hours=1))
-        city_id, weekday = "sample_city", Weekday.MONDAY
-
-        # Act
-        is_fresh = cache.is_fresh(city_id, weekday)
-
-        # Assert
-        assert not is_fresh
-
     def test_store_and_load_data(
```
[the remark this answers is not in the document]
olobuszolo: Now there are 2 seperate methods for `store` and `get`

tests/test_city_cache_data.py
RCRalph, line 1: This file should be under `tests/city_data_cache` directory

tests/test_server.py
```diff
@@ -177,10 +184,12 @@ def _assert_city_data_content(self, city_data: Any) -> None:
     @patch("tram_stop_mapper.gtfs_package.GTFSPackage.from_url")
     @patch("overpass_client.OverpassClient.get_tram_stops_and_tracks")
     @patch("overpass_client.OverpassClient.get_relations_and_stops")
-    @patch("city_data_cache.CityDataCache.is_fresh", return_value=False)
+    @patch("server.CityDataCache.get", return_value=None)
+    @patch("server.CityDataCache.store", return_value=None)
```
RCRalph, line 188:
```suggestion
    @patch("city_data_cache.CityDataCache.get", return_value=None)
    @patch("city_data_cache.CityDataCache.store", return_value=None)
```

RCRalph (commented):
Last three very minor comments, the PR looks very nice now :3

tests/city_data_cache/test_city_data_cache.py
olobuszolo: Now it is in suggested location, also renamed this file to `test_city_data_cache`

RCRalph:
@Codefident could you take a look to see if this implementation is aligned with your idea?

RCRalph (approved):
LGTM

Issue #50, opened 2025-09-16
Store cached data in zip files


RCRalph:
### Description
In order to reduce the amount of storage used by cached data, the cache for each city should be stored inside a ZIP file. This feature should be implemented using the `zipfile` library. The files should be edited and not re-created when another cache needs to be stored.

### Definition of done
Cached data for cities is stored in ZIP files instead of regular files.

PR #52, opened 2025-09-28
Store cached data in zip files

olobuszolo:
### Related issues
- #50 

### Short description
Cached data is stored in ZIP files instead of regular directories.

### DIFF
--- a/tests/city_data_cache/test_city_data_cache.py
+++ b/tests/city_data_cache/test_city_data_cache.py
@@ -90,7 +90,7 @@ class TestCityDataCache:
         krakow_zip = city_cache_dir / "krakow.zip"
 
         with ZipFile(krakow_zip) as zip_file:
-            initial_file_count = len(list(zip_file.namelist()))
+            initial_file_count = len(zip_file.namelist())
 
         # Act
         cache.store("krakow", datetime.date(2025, 9, 19), krakow_response_city_data)
@@ -125,7 +125,7 @@ class TestCityDataCache:
         cache = CityDataCache(cache_directory=city_cache_dir, max_file_count=2)
         krakow_zip = city_cache_dir / "krakow.zip"
         with ZipFile(krakow_zip) as zip_file:
-            initial_file_count = len(list(zip_file.namelist()))
+            initial_file_count = len(zip_file.namelist())
 
         # Act
         cache.store("krakow", datetime.date(2025, 9, 19), krakow_response_city_data)
@@ -133,6 +133,6 @@ class TestCityDataCache:
         # Assert
         with ZipFile(krakow_zip) as zip_file:
             dates = zip_file.namelist()
-            assert len(list(zip_file.namelist())) == 2
+            assert len(dates) == 2
             assert "2025-09-19" in dates
-            len(dates) <= initial_file_count
+            assert len(dates) <= initial_file_count


src/city_data_cache/city_data_cache.py
```diff
@@ -20,62 +23,79 @@ def __init__(
...
+        city_zip_path = self._zip_city_path(city_id)
+        file_name = f"{date.isoformat()}.json"
+
+        if not city_zip_path.exists():
             return None
 
-        return ResponseCityData.model_validate_json(
-            cache_file_path.read_text(encoding="utf-8")
-        )
+        with ZipFile(city_zip_path, "r") as zip_file:
```
RCRalph, line 36: By default files are opened in `"r"` mode:
```suggestion
        with ZipFile(city_zip_path) as zip_file:
```

src/city_data_cache/city_data_cache.py
```diff
@@ -20,62 +23,79 @@ def __init__(
...
+            if file_name not in zip_file.namelist():
+                return None
+            with zip_file.open(file_name) as file:
+                return ResponseCityData.model_validate_json(file.read().decode("utf-8"))
```
RCRalph, line 40: We can load JSON files directly from bytes (decoding to `utf-8` will be used under the hood)
```suggestion
                return ResponseCityData.model_validate_json(file.read())
```

src/city_data_cache/city_data_cache.py
```diff
@@ -20,62 +23,79 @@ def __init__(
...
-    def _get_path_to_city_cache(self, city_id: str) -> Path:
-        return self.cache_directory / city_id
+    def _zip_city_path(self, city_id: str) -> Path:
+        return self.cache_directory / f"{city_id}.zip"
 
     def get(self, city_id: str, date: datetime.date) -> ResponseCityData | None:
-        cache_file_path = self._get_path_to_cache_for_file(city_id, date)
-        if not cache_file_path.is_file():
+        city_zip_path = self._zip_city_path(city_id)
+        file_name = f"{date.isoformat()}.json"
```
RCRalph, line 31: How about we drop the file extension and just use ISO-formatted dates? It could make the implementation a bit simpler

src/city_data_cache/city_data_cache.py
RCRalph, line 1: I can see we're opening the same zip file multiple times in this file, which may hurt the performance quite a bit. Would it be possible to simplify the implementation so the ZIP file is opened only once?

src/city_data_cache/city_data_cache.py
```diff
@@ -20,62 +23,79 @@ def __init__(
...
+            return sorted(
+                [
+                    datetime.date.fromisoformat(name[:-5])
+                    for name in zip_file.namelist()
+                    if name.endswith(".json")
+                ],
+                reverse=True,
+            )
+
+    def _rebuild_zip(self, city_zip_path: Path, files_to_remove: list[str]) -> None:
```
RCRalph, line 57: I think a more appropriate argument would be `files_to_copy` instead of `files_to_remove`, I'm not quite sure why but it makes more sense to me. What do you think?

src/city_data_cache/city_data_cache.py
RCRalph: Yes, that's what I was thinking. The file content will be the same, just the file name would change.

olobuszolo: Okay, good idea

src/city_data_cache/city_data_cache.py
RCRalph: Currently you're opening the same zip file in read mode inside `_remove_redundant_files` and `_rebuld_zip`. I think you should be doable to open this file only once there.

RCRalph (commented):
Looks good, now onto the tests :3

tests/city_data_cache/test_city_data_cache.py
```diff
@@ -119,14 +123,16 @@ def test_store_remove_redundant_files(
     ) -> None:
         # Arrange
         cache = CityDataCache(cache_directory=city_cache_dir, max_file_count=2)
-        krakow_dir = city_cache_dir / "krakow"
-        initial_file_count = len(list(krakow_dir.iterdir()))
+        krakow_zip = city_cache_dir / "krakow.zip"
+        with ZipFile(krakow_zip) as zip_file:
+            initial_file_count = len(list(zip_file.namelist()))
```
RCRalph, line 128: As above:
```suggestion
            initial_file_count = len(zip_file.namelist())
```

RCRalph (approved):
Good job :3

